PR #101, opened 2021-12-08
change some #[cfg] tests to export functions on linux

rezural:
This relaxes some of the cfg tests so that relevant functions are defined on linux. This enables hotham to build at least on linux (ubuntu 20.04).

I am not able to run examples, as I do not have the openxr_loader.so file, as there don't appear to be any ubuntu packages.

Cheers


kanerogers:
Thanks so much for this, @rezural! I'll take a look over this tomorrow. 😁 

PS. Congrats on being **the very first external contributor to Hotham**!!

kanerogers (approved):

PR #132, opened 2022-01-22
Demo beginner's guide

pierre-l:
Write short instructions on how to run the demo on the Oculus Quest
Add a bash script equivalent run_on_device.ps1: run_on_device.sh


kanerogers:
Thank you so much for this @pierre-l! I've made a couple small text changes just for consistency with Hotham's _fun loving and upbeat_ brand, but otherwise this is a ⭐ ⭐ ⭐ ⭐ ⭐  PR. 

you rock

Issue #59, opened 2021-08-24
Implement Clippy


kanerogers:
## Description
Clippy will probably get mad.

## TODO
- [X] Incorporate [Clippy](https://github.com/rust-lang/rust-clippy)
- [X] Too Many Arguments problems
- [X] Non Send Fields in Send Type problem
- [X] Type Complexity Problems
- [X] Remove commented references to Clippy in code
## Problems

### [Clippy: Too Many Arguments](https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments)

1. [panel.rs](https://github.com/leetvr/hotham/blob/main/hotham/src/components/panel.rs#L79) `add_panel_to_world` (10/7) 

```rust
  --> hotham/src/components/panel.rs:79:1                                                                                                                                                                       
   |                                                                                                                                                                                                            
79 | / pub fn add_panel_to_world(                                                                                                                                                                               
80 | |     text: &str,                                                                                                                                                                                          
81 | |     width: u32,                                                                                                                                                                                          
82 | |     height: u32,                                                                                                                                                                                         
...  |                                                                                                                                                                                                          
89 | |     world: &mut World,                                                                                                                                                                                   
90 | | ) -> Entity {                                                                                                                                                                                            
   | |___________^                                                                                                                                                                                              
   |                                                                                                                                                                                                            
```

2. [gltf_loader.rs](https://github.com/leetvr/hotham/blob/main/hotham/src/gltf_loader.rs#L88) `load_node` (8/7) 

```rust
  --> hotham/src/gltf_loader.rs:88:1                                                                                                                                                                            
   |                                                                                                                                                                                                            
88 | / fn load_node(                                                                                                                                                                                            
89 | |     node_data: &gltf::Node,                                                                                                                                                                              
90 | |     gltf_buffer: &[u8],                                                                                                                                                                                  
91 | |     vulkan_context: &VulkanContext,                                                                                                                                                                      
...  |                                                                                                                                                                                                          
96 | |     images: &[gltf::image::Data],                                                                                                                                                                        
97 | | ) -> Result<()> {                                                                                                                                                                                        
   | |_______________^                                                                                                                                                                                          
   |                                                                                                                                                                                                            
```

3. [image.rs](https://github.com/leetvr/hotham/blob/main/hotham/src/image.rs#L17) `new` (8/7) 

```rust
warning: this function has too many arguments (8/7)                                                                                                                                                             
[...]

PR #137, opened 2022-02-02
Fix some clippy lints

jam1garner:
Meant to get some more handled but didn't find the time, figure I might as well push the minuscule parts I got to.


jam1garner:
See also #59

kanerogers:
Nice one @jam1garner! Thanks for this - dealing with clippy is going to be fun. :)

PR #168, opened 2022-03-10
Avoid the vec allocation in crab-saber state changes

pierre-l:
Hello,

Here is a proposal for the removal of the `objects_to_hide` and `objects_to_show` vectors in `crab-saber`'s `handle_state_change` function.

### DIFF
--- a/examples/crab-saber/src/lib.rs
+++ b/examples/crab-saber/src/lib.rs
@@ -4,7 +4,7 @@ mod systems;
 
 use hotham::{
     components::Visible,
-    hecs::World,
+    hecs::{Entity, World},
     schedule_functions::{
         apply_haptic_feedback, begin_frame, begin_pbr_renderpass, end_frame, end_pbr_renderpass,
         physics_step,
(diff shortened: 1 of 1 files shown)


examples/crab-saber/src/lib.rs
```diff
@@ -177,57 +177,54 @@ fn handle_state_change(
...
-        }
-    }
-
-    for e in objects_to_show.drain(..) {
-        world.insert_one(e, Visible {}).unwrap();
-    }
 }
 
-fn init(engine: &mut Engine) -> Result<(World, GameContext), HothamError> {
+fn init(engine: &mut Engine) -> (World, GameContext) {
```
kanerogers, line 224: Oh, that's a good point! No need to wrap this in a result!

Just to keep things clean - do you think you could split this off into a separate PR? If not, no dramas, I can take care of it.

examples/crab-saber/src/lib.rs
```diff
@@ -177,57 +177,54 @@ fn handle_state_change(
     game_context: &mut GameContext,
     world: &mut World,
 ) {
-    let mut objects_to_hide = Vec::new();
-    let mut objects_to_show = Vec::new();
-
     match (previous_state, current_state) {
         (SessionState::VISIBLE, SessionState::FOCUSED) => {
+            fn show(world: &mut World, entity: Entity) {
```
kanerogers, line 182: Do you think you could define the function outside the match statement? Might be a bit cleaner!

kanerogers (changes requested):
Looks good @pierre-l - I can't for the life of me remember why I batched these calls up like that 😆 

I've made a couple small suggestions, but overall this is looking good!

If you have some time it'd be good to get a profile of this change (as we're essentially making a perf improvement here), but otherwise I think we can safely assume that skipping needless vec allocation is a win. 👍🏻

examples/crab-saber/src/lib.rs
pierre-l: Sure, done.

kanerogers (approved):
perfect

PR #169, opened 2022-03-10
Fix all Clippy warnings and errors

pierre-l:
#168 raised CI errors caused by unrelated Clippy errors.
Here are suggestions to fix both the errors and the warnings.

I left three TODOs, these are cases that would deserve their own PRs IMHO.


kanerogers (approved):
Looks good to me @pierre-l ! Thanks so much!

For the TODOs, if you could create a separate issue (assuming they're not covered in the ones already created), that would be amazing!

PR #170, opened 2022-03-14
Implement a `Engine::run` method

pierre-l:
**Draft: needs more doc, tests, maybe a trait.**
Implement a `Engine::run` method
Make `Engine::update` private

This makes the `Engine` a bit more foolproof, preventing anyone from forgetting to call `Engine::update`.

### DIFF
--- a/examples/crab-saber/src/lib.rs
+++ b/examples/crab-saber/src/lib.rs
@@ -37,19 +37,17 @@ pub fn real_main() -> HothamResult<()> {
     let mut hotham_queries = Default::default();
     let mut crab_saber_queries = Default::default();
 
-    while let Ok((previous_state, current_state)) = engine.update() {
+    engine.run(|engine, previous_state, current_state| {
         tick(
             previous_state,
             current_state,
-            &mut engine,
+            engine,
             &mut world,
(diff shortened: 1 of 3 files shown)


kanerogers:
Hey, @pierre-l, thanks so much for this PR!

I really like the direction you're going with here, and I think this is definitely going to be something we eventually move towards - probably something similar to Bevy's App API. 

At the moment though I'm going to be a bit pedantic and just push back on this sort of big API change. What I'm really trying to do with Hotham is make it emerge as a result of building games. I'm still yet to build a game with Hotham (YET!) so I haven't got much of a feel for how the API is going to work, but I really want to make sure our future work on the project is shaped by these kinds of real world experiences.

I hope this hasn't dissuaded you (or anyone else!) from contributing to the project - I just want to emphasise again how happy it makes me to wake up in the morning and see this kind of PR. I'm still kicking myself that other people give a shit about my insane little toy VR engine!

pierre-l:
Hey @kanerogers, all good! That's pretty much the reason it was unfinished: it felt nice but maybe premature.

pierre-l:
Thanks for the feedback!

PR #171, opened 2022-03-15
Get rid of the unneeded `Result` in `crab-saber`'s `init` function

pierre-l:
`init` never fails so the `Result` can be removed.


kanerogers (approved):
Perfect

PR #183, opened 2022-04-20
Fix simulator, android debug builds.

jmgao:


kanerogers:
@jmgao Thanks so much for the contribution! I've been able to replicate the segfault locally using non NVIDIA hardware and confirmed that your fix has resolved the issue.

I haven't been able to test the Android fix but this seems like a fairly straightforward change.

kanerogers:
This has been superseded by #183 due to a minor clippy fix.

PR #197, opened 2022-04-24
Allow customization of OpenXR application name, version, required layers.

jmgao:
Make it so that applications can configure a bunch of OpenXR stuff by adding an EngineBuilder and XrContextBuilder to plumb things down to.

While we're at it, fix the android build, and make the ndk/ndk-glue versioning less restrictive so that top of tree openxr works.

### DIFF
--- a/hotham/src/engine.rs
+++ b/hotham/src/engine.rs
@@ -25,34 +25,18 @@ pub static ANDROID_LOOPER_NONBLOCKING_TIMEOUT: Duration = Duration::from_millis(
 #[cfg(target_os = "android")]
 pub static ANDROID_LOOPER_BLOCKING_TIMEOUT: Duration = Duration::from_millis(i32::MAX as _);
 
-/// The Hotham Engine
-/// A wrapper around the "external world" from the perspective of the engine, eg. renderer, XR, etc.
-/// **IMPORTANT**: make sure you call `update` each tick
-pub struct Engine {
-    should_quit: Arc<AtomicBool>,
-    #[allow(dead_code)]
-    resumed: bool,
-    event_data_buffer: EventDataBuffer,
-    /// OpenXR context
-    pub xr_context: XrContext,
-    /// Vulkan context
-    pub vulkan_context: VulkanContext,
-    /// Renderer context
-    pub render_context: RenderContext,
-    /// Physics context
-    pub physics_context: PhysicsContext,
-    /// Audio context
-    pub audio_context: AudioContext,
-    /// GUI context
-    pub gui_context: GuiContext,
-    /// Haptics context
-    pub haptic_context: HapticContext,
-}
+/// Builder for `Engine`.
+#[derive(Default)]
+pub struct EngineBuilder {}
 
-impl Engine {
-    /// Create a new instance of the engine
-    /// NOTE: only one instance may be running at any one time
+impl EngineBuilder {
+    /// Create an `EngineBuilder`
     pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Build the `Engine`
+    pub fn build(self) -> Engine {
         #[allow(unused_mut)] // Only Android mutates this.
         let mut resumed = false;
         let should_quit = Arc::new(AtomicBool::from(false));
@@ -75,10 +59,11 @@ impl Engine {
             .expect("!!FATAL ERROR - Unable to initialise renderer!");
         let gui_context = GuiContext::new(&vulkan_context);
 
-        let mut engine = Self {
+        let mut engine = Engine {
             should_quit,
             resumed,
             event_data_buffer: Default::default(),
+
             xr_context,
             vulkan_context,
             render_context,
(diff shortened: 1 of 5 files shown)


hotham/src/resources/xr_context.rs
```diff
@@ -354,45 +396,24 @@ pub(crate) fn create_xr_session(
...
-    let xr_entry = xr::Entry::load_from(path)?;
+    let xr_entry = if let Some(path) = path {
+        xr::Entry::load_from(path)?
+    } else {
+        xr::Entry::load()?
+    };
     let xr_app_info = openxr::ApplicationInfo {
-        application_name: "Hotham Asteroid",
-        application_version: 1,
+        application_name: application_name.unwrap_or("Hotham Asteroid"),
```
kanerogers, line 411: nit: Hotham Asteroid is a very old reference to the original "Asteroid" demo I created. "Hotham Application" is a better substitute.

hotham/src/engine.rs
```diff
@@ -25,34 +25,40 @@ pub static ANDROID_LOOPER_NONBLOCKING_TIMEOUT: Duration = Duration::from_millis(
...
-    /// Create a new instance of the engine
-    /// NOTE: only one instance may be running at any one time
+impl<'a> EngineBuilder<'a> {
+    /// Create an `EngineBuilder`
     pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Set the OpenXR application name
+    pub fn application_name(&mut self, name: Option<&'a str>) -> &mut Self {
```
kanerogers, line 43: Interesting - a builder pattern is definitely where I wanted to end up but I've been a bit shy about creating one. But this does feel like a reasonable start since up until this point we've not really allowed any customisation.

Vulkan also has an `application_name` field (or similar). Think you could plumb this down into `VulkanContext` as well?

hotham/src/engine.rs
```diff
@@ -25,34 +25,40 @@ pub static ANDROID_LOOPER_NONBLOCKING_TIMEOUT: Duration = Duration::from_millis(
...
+    }
+
+    /// Set the OpenXR application version
+    pub fn application_version(&mut self, version: Option<u32>) -> &mut Self {
+        self.application_version = version;
+        self
+    }
+
+    /// Set the required OpenXR extensions
+    pub fn openxr_extensions(&mut self, extensions: Option<xr::ExtensionSet>) -> &mut Self {
```
kanerogers, line 55: Nice - good call. 👍🏻

examples/crab-saber/Cargo.toml
```diff
@@ -19,7 +19,7 @@ rand = "0.8.0"
 approx = "0.5"
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk-glue = "=0.6.0"
+ndk-glue = "0.6"
```
kanerogers, line 22: There's historical reasons for this, unfortunately.

We *MUST* keep our `ndk-glue` in lockstep with the rest of the `ndk-glue` ecosystem. Several upstream packages, such as `openxr` and `cpal` will break if the `ndk-glue` version gets out of sync.

I'd prefer to leave this as is for now, and approach this by switching to `ndk-context` as documented here:

https://github.com/rust-windowing/android-ndk-rs/issues/211

kanerogers (changes requested):
This looks fantastic as always @jmgao 👍🏻 

Have left a couple of notes in the PR!

examples/crab-saber/Cargo.toml
jmgao: Doesn't cargo's semver resolution solve this? My understanding is that compatible versions of libraries (i.e. 0.6.x) are required to resolve to one specific version. The problem that this is intended to fix is that `openxr` [no longer depends on `ndk-glue`, but depends on `ndk-context` instead](https://github.com/Ralith/openxrs/commit/556b7d83cba5e7eac737bb10ea6c1b499377d66f), but an application using `hotham` can't force the version of `ndk-glue` to 0.6.2, which is the version that starts using `ndk-context`.

kanerogers: re: semver resolution, no. I thought the same thing but was unpleasantly surprised when things blew up my face on a fresh build when something in the `ndk-glue` ecosystem had been updated. To be clear, we could still get bitten by other dependencies but this is still a start.

re: `openxr` and `ndk-context` can you expand on that a bit further?

jmgao: When we update to the next version of `openxr` (or in my case, if you use `[patches.crates-io]` to move to a top of tree version to get access to `XR_FB_display_refresh_rate`), things will break, because `openxr` is removed its dependency on `ndk-glue` and added one on `ndk-context` instead.

In version 0.6.2, `ndk-glue` moves its global state into `ndk-context` and properly initializes `ndk-context`, so if a sufficiently new version of `ndk-glue` is used for initialization, crates that depend on `ndk-context` will also work. However, we're forcing the version to exactly 0.6.0, so the result is that hotham applications will have use the old `ndk-glue` that doesn't initialize `ndk-context`, and a new version of `openxr` will use an uninitialized `ndk-context` and immediately crash.

(BTW, we're *not* in lockstep with the rest of the ndk-glue ecosystem: [`cpal`] uses `0.6`, and `openxr` [used](https://github.com/Ralith/openxrs/blob/29cb8293db5ad48fd40e69e8ee23d452608b5438/openxr/Cargo.toml#L36) `0.6`, and effectively will be `>0.6.2`)

kanerogers: Completely agree @jmgao - moving to `ndk-context` is the way to go. I'd just like to approach that in a separate PR. 😄

kanerogers (changes requested):
Please remove the changes to `ndk-glue` for this PR - we'll handle that separately in #201

kanerogers (approved):
LGTM, thank you! 😀

PR #199, opened 2022-04-24
Add an empty .rustfmt.toml.

jmgao:
Make it so that if you have hotham as a submodule in an application
repository, `cargo fmt` doesn't keep going up and use rustfmt settings
from the outer repository.


kanerogers (approved):
LGTM! Good thinking!

PR #205, opened 2022-05-03
Require a newer version of ndk-glue.

### LINKED ISSUE #201: Switch to `ndk-context`
## Description
As Hotham has several direct and indirect dependencies that in turn depend upon `ndk-glue`, we should switch to the new `ndk-context` system which would allow us to avoid running into issues when our version does not match that of upstream.

## TODO
- [ ] Use `ndk-context` where appropriate

jmgao:
ndk-glue 0.6.2 moves its global state into another crate, ndk-context,
and initializes it. Other crates are starting to move to depending on
ndk-context instead, e.g. cpal depending on oboe '0.4', which is
getting resolved to a newly released 0.4.6 which depends on ndk-context.
Since we pin ndk-glue to exactly version 0.6.0, we're getting the old
ndk-glue alongside ndk-context, which remains uninitialized, leading to
explosions.

Fixes #201.

### DIFF
--- a/examples/crab-saber/Cargo.toml
+++ b/examples/crab-saber/Cargo.toml
@@ -19,7 +19,7 @@ rand = "0.8.0"
 approx = "0.5"
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk-glue = "=0.6.0"
+ndk-glue = "0.6"
 
 [package.metadata.android]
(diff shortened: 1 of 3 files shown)


kanerogers (approved):
LGTM! Thanks @jmgao

PR #210, opened 2022-05-24
Fix spelling and typos

rasmusgo:
Biggest thing is changing from the Queen’s English to Freedom English, in particular for the spelling of "color".


kanerogers:
I always new this day would come. It is with a heavy heart that I accept the supremacy of Freedom English. 🇺🇸 

Well may we say "God save the Queen", because nothing will save Noah Webster.

kanerogers (approved):

pwc:
image

PR #211, opened 2022-05-25
Fix haptic feedback when moving fast from one button to another

rasmusgo:
Remember hover status for each button to make sure that newly hovered buttons give haptic feedback even if another button was hovered last frame.


sinoth (approved):
Verified the previous incorrect behavior of haptics missing when cursor goes quickly from one button immediately to another button the next frame. This PR successfully fixes the issues and no haptics are missed.

PR #224, opened 2022-06-15
Kanerogers/issue222

kanerogers:
## Background
Adds the Sponza stress test


kanerogers:
> Some good stuff in here -- particularly the dynamic vertex buffer!

Yeah that fucker was a real treat!

PR #229, opened 2022-07-01
Applying smoother movement to the camera in the simulator

louisgjohnson:
Changes:

- added a struct to store inputs over frames, so we can apply multiple directional movement to the camera
- added last_frame_time to the state so I can smooth the camera movement

Possible Todo's:

- Move camera in rotation direction, for full 3D movement, maybe something similar to unity? 
- Might be worth adding test but probably not that much value?


kanerogers:
That is _butter smooth_, @louisgjohnson ! Nice one. 😁 

Yes, implementing a proper "fly-cam" style movement where movement is relative to the orientation (eg. if I turn 90 degrees to the left, pushing W will send me along the x axis instead of z) would be a big win! If you could create an issue for that, that would be super!

hotham-simulator/src/inputs.rs
```diff
@@ -0,0 +1,37 @@
+use std::collections::HashSet;
+
+use winit::event::{ElementState, VirtualKeyCode};
+
+#[derive(Debug, Clone)]
+pub enum HothamInputEvent {
```
kanerogers, line 6: nitpick: This type doesn't add much beyond winit's - do we really need a custom implementation?

hotham-simulator/src/state.rs
```diff
@@ -196,25 +202,38 @@ impl State {
         let mut x_rot = 0.0;
         let mut y_rot = 0.0;
 
-        match self.event_rx.as_ref()?.try_recv().ok()? {
-            HothamInputEvent::KeyboardInput { key } => match key? {
+        let now = Instant::now();
```
kanerogers, line 205: nit: a couple of comments explaining the logic could be helpful here

kanerogers (commented):
Looking good! I've left a couple of comments before merging in. 😄

louisgjohnson:
> That is _butter smooth_, @louisgjohnson ! Nice one. 😁
> 
> Yes, implementing a proper "fly-cam" style movement where movement is relative to the orientation (eg. if I turn 90 degrees to the left, pushing W will send me along the x axis instead of z) would be a big win! If you could create an issue for that, that would be super!

Haha thanks! I've also updated it now so it moves in the camera rotation direction instead of the same 4 axis each time. 

I might raise a ticket to update the mouse movement though as that seems like it could use some work but not sure how to approach it right now

hotham-simulator/src/inputs.rs
louisgjohnson: Yeah good point, much better now its removed

hotham-simulator/src/state.rs
louisgjohnson: added a couple of small comments but can always add more if you'd like

kanerogers: This is good enough for now! In general for these sorts of things a docstring above the function is nice but it's not critical here

kanerogers:
Looking good but still feels a _little_ janky in places! I think the mouse sensitivity is a bit high and I was able to end up in a situation where x axis mouse movement was translating into z rotation (eg. roll).

Camera controllers are super annoying! My go-to reference is always [the example from wgpu](https://sotrh.github.io/learn-wgpu/intermediate/tutorial12-camera/) - no shame in lifting from that!

kanerogers (changes requested):
Looking good but still feels a little janky in places! I think the mouse sensitivity is a bit high and I was able to end up in a situation where x axis mouse movement was translating into z rotation (eg. roll).

Camera controllers are super annoying! My go-to reference is always [the example from wgpu](https://sotrh.github.io/learn-wgpu/intermediate/tutorial12-camera/) - no shame in lifting from that!

louisgjohnson:
> Looking good but still feels a little janky in places! I think the mouse sensitivity is a bit high and I was able to end up in a situation where x axis mouse movement was translating into z rotation (eg. roll).
> 
> Camera controllers are super annoying! My go-to reference is always [the example from wgpu](https://sotrh.github.io/learn-wgpu/intermediate/tutorial12-camera/) - no shame in lifting from that!

 Oh nice! Cheers for the reference I’ll give that a go 😁

hotham-simulator/src/state.rs
```diff
@@ -190,43 +202,94 @@ impl State {
...
+
+        // Update Rotation
+        let orientation = &mut pose.orientation;
+
+        self.camera.yaw += x_rot * mouse_sensitivity;
+        self.camera.pitch += y_rot * mouse_sensitivity;
+
+        // I think I'm converting these two types incorrectly but it seems to work and I'm too scared to break it lol
+        let rotation: Unit<Quaternion<f32>> =
+            UnitQuaternion::from_euler_angles(self.camera.pitch, self.camera.yaw, 0f32).into();
```
louisgjohnson, line 243: from the docs for this function it specs:
```
pub fn from_euler_angles(roll: T, pitch: T, yaw: T)
```

but I'm passing it as pitch, yaw and then roll so maybe I'm transforming the nalgebra::Quaternion to openxr::Quaternionf incorrectly?

I'm terrible at linear algebra and barely understand quaternions so it could definitely be a possibility

hotham-simulator/src/state.rs
```diff
@@ -60,12 +60,22 @@ pub struct State {
     pub left_hand_space: u64,
     pub right_hand_space: u64,
     pub view_poses: Vec<Posef>,
-    pub event_rx: Option<Receiver<HothamInputEvent>>,
+    pub event_rx: Option<Receiver<DeviceEvent>>,
+    pub input_state: Inputs,
+    pub last_frame_time: Instant,
+    pub camera: Camera,
```
louisgjohnson, line 66: could put the view_pose instead this camera struct I guess?

ickk (approved):
This works excellently! 
No problems running on my ubuntu machine

changes look good!

hotham-simulator/src/state.rs
kanerogers: Yeah, that's a good point 🤔 It's _possible_ that the reason it works is because you're treating pitch/yaw/roll incorrectly elsewhere. Try taking another look!

hotham-simulator/src/state.rs
```diff
@@ -190,43 +202,94 @@ impl State {
...
+        let pose = &mut self.view_poses[0];
+
+        // Update Rotation
+        let orientation = &mut pose.orientation;
+
+        self.camera.yaw += x_rot * mouse_sensitivity;
+        self.camera.pitch += y_rot * mouse_sensitivity;
+
+        // I think I'm converting these two types incorrectly but it seems to work and I'm too scared to break it lol
+        let rotation: Unit<Quaternion<f32>> =
```
kanerogers, line 242: If you want to go between `nalgebra` and `openxr`, you can use `mint` as an intermediary to do it safely. Take a look at this function as an example:

https://github.com/leetvr/hotham/blob/519c18e47a1784a46cebbedfd93bac6697b56db3/hotham/src/util.rs#L78

kanerogers (changes requested):
That euler to quaternion conversion is definitely raising my eyebrows. 😛 Added a comment about a potentially cleaner altenative

hotham-simulator/src/state.rs
louisgjohnson: ah makes sense, wasn't sure what mint was used for

louisgjohnson:
> That euler to quaternion conversion is definitely raising my eyebrows. 😛 Added a comment about a potentially cleaner alternative

haha I did update it and it seems to work exactly the same but I agree using the properly conversion methods looks a bit more readable

kanerogers (approved):

kanerogers:
LGTM 🎉

PR #230, opened 2022-07-02
Fix audio spatialization position and velocity

rasmusgo:
This PR introduces a naming convention for transforms and coordinates. Transforms are named `<space>_from_<space>` and positions are named `<thing>_in_<space>`, eg `point_in_stage = stage_from_hand * point_in_hand`. This makes it trivial to compose transforms and spot mistakes.

One such mistake was how positions and velocities were handled in `audio.rs`, the position and velocity were specified in different coordinate spaces and it turned out both were wrong. This PR fixes that by adding the OpenXR view space so that the relative position and velocity of the view and stage reference systems can be retrieved.

There is still some room for future improvements. Two things come to mind:
1. Apply an offset to the view space to move the listener closer to the ears instead of the eyes. This can be done by passing the offset instead of an identity pose to `create_reference_space`.
2.  Use a better time than `predicted_display_time`. There is a `xr_context.instance.now()` but it complained about _"KHR_win32_convert_performance_counter_time not loaded"_ when I tried it.

I have tested running crab saber with this PR on PCVR using Quest 2 + Windows 10 + oculus airlink. It would be great if someone would test this natively on Quest 2.

### DIFF
--- a/hotham/src/systems/audio.rs
+++ b/hotham/src/systems/audio.rs
@@ -20,7 +20,7 @@ pub fn audio_system(
 ) {
     for (_, (sound_emitter, rigid_body)) in query.query_mut(world) {
         // First, where is the listener?
-        let listener_location =
+        let stage_from_listener =
             get_location_from_poses(xr_context.views[0].pose, xr_context.views[1].pose);
 
         // Get the position and velocity of the entity.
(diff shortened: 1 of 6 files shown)


hotham/src/systems/audio.rs
```diff
@@ -18,32 +18,62 @@ pub fn audio_system(
...
 
-        // Now transform the position of the entity w.r.t. the listener
-        let position = listener_location
-            .inverse_transform_point(&Point3::from(*rigid_body.translation()))
-            .into();
+        // Compute relative position and velocity
+        let relative_position_in_stage =
+            (source_position_in_stage - listener_position_in_stage).into();
+        let relative_velocity_in_stage =
+            (source_velocity_in_stage - listener_velocity_in_stage).into();
```
rasmusgo, line 63: There are lots of conversions going on here. Please let me know if you know how to make this more tidy!

kanerogers: For me this is a case where explicit and verbose beats terse and complex! Happy to leave as is

kanerogers (approved):
Fantastic work. 👍🏻

kanerogers:
@rasmusgo WRT `KHR_win32_convert_performance_counter_time`  feel free to open a separate issue for that!

PR #232, opened 2022-07-04
Fix hands don't move when only one controller is connected.

### LINKED ISSUE #179: [Bug]: Simple scene hands don't move if only one controller is connected
### What happened?

Both hands seem stuck to the floor and do not receive position updates if only one Touch controller is connected. Once both are connected, the hands spring to life.

### Version

0.1.1 (Default)

### In which Hotham component are you seeing the problem on?

Example

### What VR System are you seeing the problem on?

Quest-2

### What OS are you seeing the problem on?

Windows

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ickk:
Just skip hands with invalid spaces instead of returning early.

Fixes #179.

### DIFF
--- a/hotham/src/systems/hands.rs
+++ b/hotham/src/systems/hands.rs
@@ -34,7 +34,7 @@ pub fn hands_system(
 
         // Check it's valid before using it
         if !is_space_valid(&space) {
-            return;
+            continue;
         }
 
         let pose = space.pose;


kanerogers (approved):
Nice one! 😁

PR #242, opened 2022-07-06
Allow asset importer to import models with no materials

### LINKED ISSUE #241: [Bug]: The asset importer can't import models without materials
### What happened?

If you try to `load_models_from_glb` on a .glb file with no materials then you hit [this expect](https://github.com/leetvr/hotham/blob/64f022677f1ba699387ebb2947ebe9f3ba35873a/hotham/src/rendering/primitive.rs#L131).
```
thread 'main' panicked at 'Primitives without materials are not yet supported!', hotham/src/rendering/primitive.rs:131:14
```

### Version

latest

ickk:
Reserve index 0 in the material buffer for the default material, and use that when there are no materials in the gltf file.

Fixes #241

### DIFF
--- a/hotham/src/asset_importer/mod.rs
+++ b/hotham/src/asset_importer/mod.rs
@@ -335,6 +335,8 @@ pub fn add_model_to_world(
     Some(new_root_entity)
 }
 
+// These tests are disabled for other platforms
+// https://github.com/leetvr/hotham/issues/240
 #[cfg(target_os = "windows")]
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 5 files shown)


hotham/src/asset_importer/mod.rs
```diff
@@ -436,6 +438,13 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_load_model_with_no_material() {
```
kanerogers, line 442: naice

kanerogers (approved):
LGTM!

PR #244, opened 2022-07-06
Add and call update_listener_rotation function

rasmusgo:
This is a follow up on #230.

### DIFF
--- a/hotham/src/resources/audio_context.rs
+++ b/hotham/src/resources/audio_context.rs
@@ -153,6 +153,10 @@ impl AudioContext {
         };
     }
 
+    pub(crate) fn update_listener_rotation(&mut self, rotation: mint::Quaternion<f32>) {
+        self.scene_handle.control().set_listener_rotation(rotation);
+    }
+
     /// Add a music track
     pub fn add_music_track(&mut self, mp3_bytes: Vec<u8>) -> MusicTrack {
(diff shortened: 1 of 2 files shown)


hotham/src/resources/audio_context.rs
```diff
@@ -153,6 +153,10 @@ impl AudioContext {
         };
     }
 
+    pub(crate) fn update_listener_rotation(&mut self, rotation: mint::Quaternion<f32>) {
```
kanerogers, line 156: nit: Any reason why this couldn't just be public?

hotham/src/systems/audio.rs
```diff
@@ -19,18 +20,15 @@ pub fn audio_system(
...
             &xr_context.stage_space,
             xr_context.frame_state.predicted_display_time, // TODO: Use "now" instead.
         )
         .unwrap();
 
-    if !listener_position_in_stage
-        .location_flags
-        .contains(SpaceLocationFlags::POSITION_VALID)
-    {
+    if !is_space_valid(&stage_from_listener) {
```
kanerogers, line 31: Good catch

kanerogers (approved):
LGTM, minor nitpick!

hotham/src/resources/audio_context.rs
kanerogers: Yep, good call.

PR #248, opened 2022-07-11
Add `Grabbable` marker component

### LINKED ISSUE #247: Not everything should be grabbable by the hands
Pretty sure this is the cause of a crash when the hands get close together, but it's also just a problem if the hands happen to be on top of multiple entities with colliders when gripping - as whichever entity happens to be seen by the grabbing system first will get grabbed, even if it's not actually intended to be interacted with.

Solution: Add a `Grabbable` marker component

ickk:
closes #247

### DIFF
--- a/hotham/src/systems/grabbing.rs
+++ b/hotham/src/systems/grabbing.rs
@@ -126,7 +126,22 @@ mod tests {
             &mut rigid_body_query,
         );
 
-        let hand = world.get_mut::<Hand>(hand_entity).unwrap();
+        let mut hand = world.get_mut::<Hand>(hand_entity).unwrap();
+        assert!(hand.grabbed_entity.is_none());
+
+        // Make sure hand can't grip colliders *without* a Grabbable component
+        hand.grip_value = 1.0;
+        drop(hand);
+        world.remove::<(Grabbable,)>(grabbed_entity).unwrap();
+
+        schedule(
+            &mut query,
+            &mut world,
+            &mut physics_context,
+            &mut rigid_body_query,
+        );
+
+        let hand = world.get::<Hand>(hand_entity).unwrap();
         assert!(hand.grabbed_entity.is_none());
     }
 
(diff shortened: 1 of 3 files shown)


hotham/src/systems/grabbing.rs
```diff
@@ -69,10 +73,13 @@ mod tests {
 
         let grabbed_collider = ColliderBuilder::cuboid(1.0, 1.0, 1.0).build();
         let grabbed_rigid_body = RigidBodyBuilder::new_dynamic().build();
-        let grabbed_entity = world.spawn((Info {
-            name: "Test entity".to_string(),
-            node_id: 0,
-        },));
+        let grabbed_entity = world.spawn((
```
kanerogers, line 76: Nice! Can you write a test confirming the inverse? (eg. objects without grabbable can't be grabbed)?

kanerogers (changes requested):

hotham/src/systems/grabbing.rs
```diff
@@ -119,7 +126,22 @@ mod tests {
...
+        world.remove::<(Grabbable,)>(grabbed_entity).unwrap();
+
+        schedule(
+            &mut query,
+            &mut world,
+            &mut physics_context,
+            &mut rigid_body_query,
+        );
+
+        let hand = world.get::<Hand>(hand_entity).unwrap();
```
kanerogers, line 144: Perfect!

hotham/src/systems/grabbing.rs
```diff
@@ -119,7 +126,22 @@ mod tests {
...
         );
 
-        let hand = world.get_mut::<Hand>(hand_entity).unwrap();
+        let mut hand = world.get_mut::<Hand>(hand_entity).unwrap();
+        assert!(hand.grabbed_entity.is_none());
+
+        // Make sure hand can't grip colliders *without* a Grabbable component
+        hand.grip_value = 1.0;
+        drop(hand);
+        world.remove::<(Grabbable,)>(grabbed_entity).unwrap();
```
kanerogers, line 135: nit: you should be able to just do `world.remove::<Grabbable>`, but this isn't a blocker

kanerogers (approved):
LGTM!

PR #249, opened 2022-07-11
Kanerogers/issue237

### LINKED ISSUE #237: Texture compression

kanerogers:
## What is this related to?
This PR closes #237

## What changed?
- Made significant improvements to PBR shader, including removing entirely superfluous sRGB conversions and gamma correction
- Implemented texture compression and channel swizzling to remove texture size in GPU memory
 
In general, this is a *significant* performance win.


hotham/build.rs
```diff
@@ -0,0 +1,5 @@
+pub fn main() {
+    // On Android, we must ensure that we're dynamically linking against the C++ standard library.
+    // For more details, see https://github.com/rust-windowing/android-ndk-rs/issues/167
+    println!("cargo:rustc-link-lib=dylib=c++");
```
ickk, line 4: This flag causes builds to fail on linux

ickk (changes requested):
This worked really well on my quest 2! I think I had a couple hundred helmets in the stress test before I got bored lol.

That said, the simulator complains about validation:
```
[HOTHAM_POLL_EVENT] State is now READY
[HOTHAM_POLL_EVENT] State is now SYNCHRONIZED
[HOTHAM_POLL_EVENT] State is now FOCUSED
[HOTHAM_TEXTURE] Importing ktx2 texture in ASTC_8X8_SRGB_BLOCK format
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ERROR_VALIDATION_FAILED_EXT', hotham/src/rendering/texture.rs:45:14
```
Is this just the validation layers doing their thing/can I make this not a panic?

Other than that if we can conditionally enable that build flag only for android that would be great! Otherwise I'm pretty happy, nice work!

benchmarks/stress-test/src/lib.rs
```diff
@@ -120,8 +120,9 @@ fn init(engine: &mut Engine, test: &StressTest) -> (World, HashMap<String, World
             models
         }
         StressTest::ManyHelmets => {
-            let glb_buffers: Vec<&[u8]> =
-                vec![include_bytes!("../../../test_assets/damaged_helmet.glb")];
+            let glb_buffers: Vec<&[u8]> = vec![include_bytes!(
```
louisgjohnson, line 123: just curious why does this need to be wrapped in a Vec?

PR #262, opened 2022-07-12
Fix swizzled textures on desktop

### LINKED ISSUE #261: [Rendering] Swizzled textures don't work correctly on desktop
## Background
Trying to use the swizzled, optimized textures looks like this in the simulator:
image

That's not, you know, great.

## Solutions
- [ ] Find a swizzle in the PBR shader that works on both mobile and desktop

OR 

- [ ] Swizzle the textures on import on desktop

kanerogers:
## What is this related to?
Fixes #261

## What changed?
- Squisher attempts to swizzle textures into the most optimal format
- This creates issues on desktop where we're using uncompressed textures
- To resolve this we change the swizzle on the image view when using uncompressed textures

Scenes should look correct when using uncompressed textures.


ickk (approved):
Nice!

Issue #263, opened 2022-07-12
[Renderer] GPU crash - IOCTL_KGSL_GPU_COMMAND


kanerogers:
## Background
Shared memory between CPU and GPU needs to be handled carefully:

```
07-12 18:40:02.018 11598 11675 I VrApi   : FPS=62/72,Prd=40ms,Tear=0,Early=0,Stale=20,VSnc=0,Lat=-1,Fov=0,CPU4/GPU=2/3,1171/490MHz,OC=FF,TA=0/0/0,SP=N/N/N,Mem=1804MHz,Free=560MB,PLS=0,Temp=31.0C/0.0C,TW=1.05ms,App=13.52ms,GD=0.00ms,CPU&GPU=20.85ms,LCnt=1,GPU%=0.96,CPU%=0.07(W0.10),DSF=1.00
07-12 18:40:02.069 11598 11624 I RustStdoutStderr: Adding Damaged Helmet to world
07-12 18:40:02.069 11598 11624 I RustStdoutStderr: [HOTHAM_STRESS_TEST] There are now 237 models
07-12 18:40:02.400 11598 11625 W Adreno-GSL: <gsl_ldd_control:556>: ioctl fd 56 code 0xc040094a (IOCTL_KGSL_GPU_COMMAND) failed: errno 71 Protocol error
07-12 18:40:02.400 11598 11625 W Adreno-GSL: <log_gpu_snapshot:464>: panel.gpuSnapshotPath is not set.not generating user snapshot
07-12 18:40:02.458 11598 11625 W Adreno-GSL: <gsl_ldd_control:556>: ioctl fd 56 code 0xc040094a (IOCTL_KGSL_GPU_COMMAND) failed: errno 35 Resource deadlock would occur
07-12 18:40:02.458 11598 11625 W Adreno-GSL: <log_gpu_snapshot:464>: panel.gpuSnapshotPath is not set.not generating user snapshot
```

kanerogers:
Work done so far:

- Created a separate buffer for all resources that are written by the CPU / read by the GPU each frame.
- Added an explicit pipeline barrier in the compute stage

kanerogers:
I'm still seeing this issue, which is pretty frustrating. My next approach is to use the `arm best practices` layer and keep digging through any other optimisation problems.

kanerogers:
Closed in #266

kanerogers:
As an update, we've reached out to Meta (Oculus) re: this issue and they're looking into it.

[...]

PR #264, opened 2022-07-12
Fix frustrum culling

### LINKED ISSUE #238: [Bug] Frustrum culling causes GPU crashes
## Background
When attempting to use compute based frustrum culling, we get a GPU crash. It's probably because we're very lazily re-using most of the synchronisation primitives as the graphics pipeline.

kanerogers:
## What is this related to?
Closes #238.

## What changed?
We now do compute based frustrum culling

Known issues:
- Culling begins to cause deadlocks at around 200 objects
- This will be resolved in #263

### DIFF
--- a/hotham/src/rendering/buffer.rs
+++ b/hotham/src/rendering/buffer.rs
@@ -50,7 +50,12 @@ impl<T: Sized> Buffer<T> {
 
         // Map memory
         let memory_address = device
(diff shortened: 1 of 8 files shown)


ickk (approved):
Looks good

ickk:
I got deadlocks at around 160-170

PR #278, opened 2022-07-23
Fix swizzling on desktop

### LINKED ISSUE #277: [Bug]: Normalsmaps are broken on desktop
### What happened?

Normalmaps are not working correctly when running on desktop. Here is a screenshot from simple scene example with some extra spheres added and the shader modified to output the final normals:
OculusScreenshot1658524921
 
This happens both in the simulator and running on the quest 2 via oculus air link.

The bug was not present before #249 was merged which introduced swizzling. The swizzled format is using green and alpha channels to store x and y of normals. The green channel is centered around 0.5 as expected but the alpha channel is much brighter. I think this could be caused by sRGB conversions that treat RGB different from the alpha channel.

### Version

latest

### In which Hotham component are you seeing the problem on?

Example

### What VR System are you seeing the problem on?

_No response_

### What OS are you seeing the problem on?

Windows

### Relevant log output

_No response_

rasmusgo:
This resolves #277. It turns out the issue was trivial. The blue channel (`normal.z`) was passed to the shader instead of the green channel (`normal.y`). I also took the liberty of changing the render debug mode for normals to render the computed normal instead of `inNormal.xy` or raw texture.

### Normals before
Screenshot 2022-07-23 153527

### Normals after
Screenshot 2022-07-23 153610

### Render before
Screenshot 2022-07-23 154122

### Render after
Screenshot 2022-07-23 154044

### DIFF
--- a/hotham/src/rendering/texture.rs
+++ b/hotham/src/rendering/texture.rs
@@ -244,7 +244,7 @@ fn get_component_mapping(
             r: vk::ComponentSwizzle::ZERO,
             g: vk::ComponentSwizzle::R,
             b: vk::ComponentSwizzle::ZERO,
-            a: vk::ComponentSwizzle::B,
+            a: vk::ComponentSwizzle::G,
         },
         (true, TextureUsage::MetallicRoughness) => vk::ComponentMapping {
             r: vk::ComponentSwizzle::ZERO,
--- a/hotham/src/shaders/pbr.frag
+++ b/hotham/src/shaders/pbr.frag
@@ -325,8 +325,7 @@ void main()
 				outColor.rgba = baseColor;
 				break;
 			case 2:
-				outColor.rg = (material.normalTextureID == NO_TEXTURE) ? normalize(inNormal).xy : texture(textures[material.normalTextureID], inUV).ga;
-				outColor.b = 0.0;
+				outColor.rgb = n * 0.5 + 0.5;
 				break;
 			case 3:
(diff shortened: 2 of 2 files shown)


kanerogers:
@rasmusgo is the MVP 🏆

hotham/src/shaders/pbr.frag
```diff
@@ -325,8 +325,7 @@ void main()
 				outColor.rgba = baseColor;
 				break;
 			case 2:
-				outColor.rg = (material.normalTextureID == NO_TEXTURE) ? normalize(inNormal).xy : texture(textures[material.normalTextureID], inUV).ga;
-				outColor.b = 0.0;
+				outColor.rgb = n * 0.5 + 0.5;
```
kanerogers, line 328: Fantastic idea. Sasche originally intended this to preview the textures themselves but for normals this is in fact more helpful.

hotham/src/rendering/texture.rs
```diff
@@ -244,7 +244,7 @@ fn get_component_mapping(
             r: vk::ComponentSwizzle::ZERO,
             g: vk::ComponentSwizzle::R,
             b: vk::ComponentSwizzle::ZERO,
-            a: vk::ComponentSwizzle::B,
+            a: vk::ComponentSwizzle::G,
```
kanerogers, line 247: facepalm

kanerogers (approved):
Brilliant work as always, Herr Rasmus

Issue #275, opened 2022-07-22
[Rendering] Fix frustum culling


kanerogers:
At present our frustum culling is both slow and incorrect.

An optimisation was made by @kanerogers to create a virtual frustum incorporating both cameras, but this has not proven to be correct. In addition, the underlying isVisible function in the compute shader is extremely expensive, and should, if possible, be reduced to a plane intersection check.

Ralith:
I'm assuming that #280 addressed the correctness issues. The current code looks reasonable to me in that respect, though I haven't independently audited the math used to extract clipping planes. However, one serious design issue remains:

https://github.com/leetvr/hotham/blob/c3fc0142a9e6d1f9a68d9cbe2ffd8e9def27c19c/hotham/src/resources/render_context.rs#L284-L287

This forces a full pipeline stall and host round-trip between culling and whatever operations follow (presumably, drawing). It should be replaced with appropriate pipeline barriers.

kanerogers:
Thanks for looking into this!

re: the hard wait on the Fence, how would we ensure the dispatch is complete when the CPU then reads the cull buffer? 

Noting that we are not doing multiDrawIndirect; we are simply iterating through the results of the cull buffer and guiding up our draw commands CPU side. Previously we simply used barriers and drawIndirect, which though technically correct, has serious implementation issues on the Adreno 650 hardware we're targeting.

Ralith:
Ah, I didn't realize you had resorted to reading back the results on the CPU! In that case it might be better to do the culling on the CPU exclusively rather than synchronously shuffling data back and forth, though it's hard to say for sure without profiling and/or benchmarking on the target hardware with representative scenes. Another option is conditional rendering, but that doesn't seem to be supported on the Quest 2.

Is the vendor working on fixing their bugs? Might make sense to hold off on further iteration regardless if a fix is in flight.

kanerogers:
I have thought about doing everything on the CPU, but the overhead for CPU
<> GPU seems fairly minimal from my profiling (remembering that we’ve got
unified coherent memory). At the very least, the cost of parallelism should
be in our favour for any real world number of objects.

re: vendor bugs it doesn’t seem to be so much of a bug as a design flaw.
The entire Adreno series seems to not have good support for drawIndirect,
and speaking with Meta and other engine folks (eg. Godot), the solution is
just to not even bother. We can get pretty good results with CPU batching
and GPU instancing, so I’m reasonably happy with this architecture for now,
[...]

PR #280, opened 2022-07-24
Improve culling and simulator responsiveness

rasmusgo:
* Handle all pending input events in `update_camera`- This fixes laggy input in the simulator when using a mouse with high refresh rate.
* Let the second camera follow the first. - This makes the interpolated pose culling work for the simulator.
* Replace shared camera frustum with separate frustums for the two cameras. - This removes the interpolated pose altogether.
* Move some redundant computations in `culling.comp` to the CPU so that the GPU code can be simplified.

### DIFF
--- a/hotham-simulator/src/state.rs
+++ b/hotham-simulator/src/state.rs
@@ -297,6 +297,7 @@ impl State {
         // self.spaces.get_mut(&right_hand).unwrap().position.z += z_delta;
         // self.spaces.get_mut(&right_hand).unwrap().position.x += x_delta;
 
+        self.view_poses[1] = self.view_poses[0];
         Some(())
     }
 }
--- a/hotham-simulator/src/state.rs
+++ b/hotham-simulator/src/state.rs
@@ -218,7 +218,7 @@ impl State {
         let movement_speed = 2f32 * dt;
         let mouse_sensitivity = 4f32 * dt;
 
-        if let Ok(input_event) = self.event_rx.as_ref()?.try_recv() {
+        while let Ok(input_event) = self.event_rx.as_ref()?.try_recv() {
             match input_event {
                 DeviceEvent::Key(keyboard_input) => self.input_state.process_event(keyboard_input),
                 DeviceEvent::MouseMotion { delta: (x, y) } => {
--- a/hotham/src/resources/render_context.rs
+++ b/hotham/src/resources/render_context.rs
@@ -256,16 +256,12 @@ impl RenderContext {
             left_clip_planes: Matrix4::<_>::from_rows(&[
                 normalize_plane(left_frustum_from_world.row(3) + left_frustum_from_world.row(0)),
                 normalize_plane(left_frustum_from_world.row(3) - left_frustum_from_world.row(0)),
-                // normalize_plane(left_frustum_from_world.row(3) + left_frustum_from_world.row(1)),
-                // normalize_plane(left_frustum_from_world.row(3) - left_frustum_from_world.row(1)),
                 normalize_plane(left_frustum_from_world.row(3) + left_frustum_from_world.row(2)),
                 normalize_plane(left_frustum_from_world.row(3) - left_frustum_from_world.row(2)),
             ]),
             right_clip_planes: Matrix4::<_>::from_rows(&[
                 normalize_plane(right_frustum_from_world.row(3) + right_frustum_from_world.row(0)),
                 normalize_plane(right_frustum_from_world.row(3) - right_frustum_from_world.row(0)),
-                // normalize_plane(right_frustum_from_world.row(3) + right_frustum_from_world.row(1)),
-                // normalize_plane(right_frustum_from_world.row(3) - right_frustum_from_world.row(1)),
                 normalize_plane(right_frustum_from_world.row(3) + right_frustum_from_world.row(2)),
                 normalize_plane(right_frustum_from_world.row(3) - right_frustum_from_world.row(2)),
             ]),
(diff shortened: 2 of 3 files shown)


rasmusgo:
I think this should resolve https://github.com/leetvr/hotham/issues/275.

hotham-simulator/src/state.rs
```diff
@@ -218,7 +218,7 @@ impl State {
         let movement_speed = 2f32 * dt;
         let mouse_sensitivity = 4f32 * dt;
 
-        if let Ok(input_event) = self.event_rx.as_ref()?.try_recv() {
+        while let Ok(input_event) = self.event_rx.as_ref()?.try_recv() {
```
kanerogers, line 221: Good call

hotham-simulator/src/state.rs
```diff
@@ -297,6 +297,7 @@ impl State {
         // self.spaces.get_mut(&right_hand).unwrap().position.z += z_delta;
         // self.spaces.get_mut(&right_hand).unwrap().position.x += x_delta;
 
+        self.view_poses[1] = self.view_poses[0];
```
kanerogers, line 300: Nice

hotham/src/resources/render_context.rs
```diff
@@ -229,26 +229,46 @@ impl RenderContext {
...
+        // We are only using four planes per camera. The near and far planes are not used.
+        // This link points to a paper describing the math behind these expressions:
+        // https://www.gamedevs.org/uploads/fast-extraction-viewing-frustum-planes-from-world-view-projection-matrix.pdf
+        let normalize_plane = |p: RowVector4<_>| p / p.columns(0, 3).norm();
         let cull_data = CullData {
-            frustum: frustum_projection * frustum_view,
+            clip_planes1: Matrix4::<_>::from_rows(&[
+                normalize_plane(left_frustum_from_world.row(3) + left_frustum_from_world.row(0)),
+                normalize_plane(left_frustum_from_world.row(3) - left_frustum_from_world.row(0)),
+                // normalize_plane(left_frustum_from_world.row(3) + left_frustum_from_world.row(1)),
```
kanerogers, line 259: Could you remove the commented out code?

hotham/src/resources/render_context.rs
```diff
@@ -229,26 +229,46 @@ impl RenderContext {
...
+        let right_frustum_from_world = right_frustum_from_right_camera * right_camera_from_world;
 
+        // Normals of the clipping planes are pointing towards the inside of the frustum.
+        // We are only using four planes per camera. The near and far planes are not used.
+        // This link points to a paper describing the math behind these expressions:
+        // https://www.gamedevs.org/uploads/fast-extraction-viewing-frustum-planes-from-world-view-projection-matrix.pdf
+        let normalize_plane = |p: RowVector4<_>| p / p.columns(0, 3).norm();
         let cull_data = CullData {
-            frustum: frustum_projection * frustum_view,
+            clip_planes1: Matrix4::<_>::from_rows(&[
```
kanerogers, line 256: I'd prefer these be called something like `left_clip_plane` and `right_clip_plane`, just to be consistent with the other variables, both here and in the compute shader

hotham/src/resources/render_context.rs
```diff
@@ -729,8 +749,9 @@ fn create_pipeline_layout(
 #[repr(C)]
 #[derive(Debug, Clone)]
 pub struct CullData {
-    /// View-Projection matrices (one per eye)
-    pub frustum: Matrix4<f32>,
+    /// Four clip planes per camera, one plane per row.
+    pub clip_planes1: Matrix4<f32>,
```
kanerogers, line 753: see above re: variable names

hotham/src/shaders/culling.comp
```diff
@@ -27,48 +27,24 @@ layout(std430, set = 0, binding = 1) writeonly buffer DrawCommandsBuffer {
...
-    bool isVisible = check_is_visible(cullData.frustum, center, radius);
+    vec4 center4 = vec4(d.boundingSphere.xyz, 1);
+    vec4 negRadius4 = -d.boundingSphere.wwww;
+
+    // Treat as visible if at least one eye may see it
+    bool isVisible = any(bvec2(
+        all(greaterThan(cullData.clipPlanes1 * center4, negRadius4)),
+        all(greaterThan(cullData.clipPlanes2 * center4, negRadius4))
+    ));
     drawCommandsBuffer.drawCommands[id].instanceCount = isVisible ? 1 : 0;
```
kanerogers, line 49: Minor optimisation: cast the bool to a uint with uint(isVisible) - saves a couple instructions

kanerogers (changes requested):
Overall this looks good! Just some minor changes suggested. :)

In future it would be nice to keep PRs like these separate (eg. the simulator responsiveness) but I understand how annoying that can be for small fixes

hotham/src/resources/render_context.rs
rasmusgo: I left them in as a trace of what was omitted but it's a bit redundant given the comment above.

kanerogers (approved):
LGTM! Fantastic work as always @rasmusgo

hotham/src/shaders/culling.comp
Ralith: FWIW, there's approximately zero chance the driver won't produce equivalent code for those.

kanerogers: On desktop land these assumptions are reasonable, but Qualcomm's documentation constantly warns about not assuming the driver's compiler will do what you want. 😛

PR #281, opened 2022-07-25
Use 'Q' or 'Escape' to exit the simulator

pwc:
Pressing the “X” on the simulator window controls doesn’t seem to do anything. Allow the user to exit using either Escape (logical) or Q (easy to press).

Questions:
(1) Should the “X” work? i.e. is this Linux/X11/KDE’s fault?
(2) Am I setting up the quittey thing the right way? I just copied it from `request_exit_session`.


rasmusgo:
Pressing the "X" works for me on windows. We should make that work on linux too!

kanerogers:
What technically happens on Windows at least is closing the window causes the renderer to crash, which brings down the rest of the simulator. 😄 

If someone can be bothered, setting "should quit" or similar would be pretty helpful. But in the meantime, totally happy to merge this in to keep the Linux people happy!

kanerogers (approved):
Works nicely for me on Windows! Closing. :)

PR #282, opened 2022-07-25
Fix build by fixing ndk-glue version

rasmusgo:
There is a newly released ndk version 0.7.0 that sneaks in and causes build issues. This is an attempt to resolve the build problems until we choose to upgrade.


kanerogers (approved):
LGTM!

PR #288, opened 2022-07-26
Added crate dolly for handling camera transform/rotation

louisgjohnson:
Introducing the crate([dolly](https://github.com/h3r2tic/dolly)) into hotham-simulator for handling the camera transform and rotation


hotham-simulator/src/state.rs
```diff
@@ -216,7 +204,7 @@ impl State {
 
         let dt = delta.as_secs_f32();
         let movement_speed = 2f32 * dt;
-        let mouse_sensitivity = 4f32 * dt;
+        let mouse_sensitivity = 40f32 * dt;
```
louisgjohnson: not sure why but I had to massive bump the mouse_sensitivity up

rasmusgo: `dt` should not be here because we don't want our framerate to decide how fast the mouse turns the view.

louisgjohnson: yeah good point, we probably don't really need it for camera directional movement either

rasmusgo:
It seems like dolly is overkill for what we are asking it to do at the moment. We are not using any smoothing or putting the camera on an arm. Do you have any particular features in mind to motivate adding dolly (and indirectly glam)?

kanerogers:
Ah hrm, that is a good point @rasmusgo , I didn't realise how opinionated `dolly` was (using another maths library is not good). 😦 

I'm really sorry I didn't look into `dolly` more before I suggested it! I don't want you to feel like you wasted your time on this @louisgjohnson - what do you think `dolly` gives us that we don't already have?

kanerogers:
I'm going to close this one off for now! Sorry about that @louisgjohnson 😦

PR #292, opened 2022-07-27
Fix bounding sphere transformations

rasmusgo:
~~This follows #291.~~

Use `GlobalTransform` instead of `LocalTransform` for computing transformed bounding spheres. This takes the full transformation from local to world into account and properly deals with hierarchies of parents and children.

### DIFF
--- a/hotham/src/rendering/primitive.rs
+++ b/hotham/src/rendering/primitive.rs
@@ -143,7 +143,13 @@ impl Primitive {
         let center_in_local: Point3<_> = self.bounding_sphere.xyz().into();
         let center_in_world =
             Point3::<_>::from_homogeneous(transform.0 * center_in_local.to_homogeneous()).unwrap();
+
+        // The linear part contains the rotation and scale, we are interested in the scale.
         let world_from_local_linear_part = transform.0.fixed_slice::<3, 3>(0, 0);
+
+        // The scale of the sphere is taken as the largest scale in any dimension.
+        // If the scale is non-uniform, there could be a tighter bounding sphere.
+        // If the scale is uniform, the quality of the bounding sphere will be unchanged.
         let scale = world_from_local_linear_part
             .column(0)
             .magnitude_squared()
--- a/hotham/src/rendering/primitive.rs
+++ b/hotham/src/rendering/primitive.rs
@@ -1,11 +1,11 @@
 use crate::{
     asset_importer::ImportContext,
-    components::Transform,
+    components::TransformMatrix,
     rendering::{material::NO_MATERIAL, vertex::Vertex},
     resources::render_context,
 };
 use itertools::izip;
-use nalgebra::{vector, Vector3, Vector4};
+use nalgebra::{vector, Point3, Vector3, Vector4};
 use render_context::RenderContext;
 
 /// Geometry for a mesh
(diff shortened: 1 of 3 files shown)


kanerogers (approved):
Not only is this more correct, it's *even faster*. Bravo, @rasmusgo

kanerogers:
Unfortunately I can't merge this in yet as it contains #291. If you're able to perhaps just rebase on main with _only_ `get_bounding_sphere`, I'll merge this in ASAP

hotham/src/rendering/primitive.rs
```diff
@@ -139,14 +139,26 @@ impl Primitive {
...
     /// Get a bounding sphere for the primitive, applying a transform
-    pub fn get_bounding_sphere(&self, transform: &Transform) -> Vector4<f32> {
-        let mut center = self.bounding_sphere.xyz();
-        let mut radius = self.bounding_sphere.w;
-
-        center += transform.translation;
-        radius *= transform.scale.max();
-
-        [center.x, center.y, center.z, radius].into()
+    pub fn get_bounding_sphere(&self, transform: &GlobalTransform) -> Vector4<f32> {
```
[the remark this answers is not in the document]
rasmusgo: Erhm, this is straight from my brains. I'll try to explain how it works.

rasmusgo:
If "Remove Transform..." is causing problems it should revert cleanly.

rasmusgo:
So is this correct? Why? Or more exactly, how can we know that all points inside the bounding sphere in local space are contained in this new sphere when transformed into world space?

If this holds:

$$|p_{local} - c_{local}| \le r_{local}$$

then we want this to hold:

$$|p_{world} - c_{world}| \le r_{world}$$

These are related by a transformation that may contain rotation, translation and scale, aka an [affine transformation](https://en.wikipedia.org/wiki/Affine_transformation):

$$p_{world} = T_{world \leftarrow local} * p_{local}$$

We pick the new sphere center as the transformed center of the bounding sphere in local space:

$$c_{world} = T_{world \leftarrow local} * c_{local}$$

Because the sphere center is transformed in the same way as the points, a point at the center of the sphere in local space will also be in the center of the new sphere when transformed into world space.

So what about other points? We can simplify the problem by removing the translational part of the transformation because of how the new center is chosen. This leaves us with rotation and scale. This is the upper left 3x3 part of the 4x4 transformation matrix.

$$p_{world} = T_{world \leftarrow local} * p_{local} = L_{world \leftarrow local} * p_{local} + t_{world \leftarrow local}$$
$$p_{world} - c_{world}  = (L_{world \leftarrow local} * p_{local} + t_{world \leftarrow local}) -(L_{world \leftarrow local} * c_{local} + t_{world \leftarrow local})$$
$$p_{world} - c_{world}  = L_{world \leftarrow local} * (p_{local} - c_{local})$$

Let's insert that into the inequality that we want to hold:

$$|p_{world} - c_{world}| \le r_{world}$$
$$|L_{world \leftarrow local} * (p_{local} - c_{local})| \le r_{world}$$


Now we bring out the sledge hammer of linear algebra: The [Singular Value Decomposition (SVD)](https://en.wikipedia.org/wiki/Singular_value_decomposition). Every matrix can be written as a product of three special matrices: rotation, axis aligned scale, and rotation. This is super useful for proving stuff because these special matrices have lots of special properties that can be used. If we would do this to the linear part of the transformation, we get:

$$L_{world \leftarrow local} = UDV'$$
$$|UDV' * (p_{local} - c_{local})| \le r_{world}$$

U and V are rotation matrices. Their columns all have unit length and are orthogonal to each other, same goes for their rows. Their transpose is also their inverse. The matrix D contains the scale and is a diagonal matrix. Multiplication with a diagonal matrix scales the rows if applied from the left or scales the columns if applied from the right.

Rotating a vector does not change its length so if we pick the new radius as the local radius times the biggest element in the diagonal matrix D, we will get a new sphere in world space that is big enough.

$$|UDV' * (p_{local} - c_{local})| \le max(D) * |p_{local} - c_{local}| = max(D) * r_{local}$$
$$r_{world} = max(D) * r_{local}$$


Ok, great! **But that is not what the code says.**

Admittedly, it does not. It scales the radius by the magnitude of the largest column of the linear part of the transformation matrix. However, if the scaling is axis aligned in local space, the matrix V will be an identity matrix representing zero rotation and we are left with $UD$. Multiplying on the right with a diagonal matrix scales the columns. The columns have unit length to start with. This means that **the code works as long as the scale is uniform or axis aligned in local space**. If this is not the case, it may underestimate the radius.

hotham/src/rendering/primitive.rs
```diff
@@ -139,14 +139,32 @@ impl Primitive {
...
+
+        // The scale of the sphere is taken as the largest scale in any dimension.
+        // If the scale is non-uniform, there could be a tighter bounding sphere.
+        // If the scale is uniform, the quality of the bounding sphere will be unchanged.
+        let scale = world_from_local_linear_part
+            .column(0)
+            .magnitude_squared()
+            .max(world_from_local_linear_part.column(1).magnitude_squared())
+            .max(world_from_local_linear_part.column(2).magnitude_squared())
+            .sqrt();
```
rasmusgo, line 158: This can be replaced with an SDV if we want to handle the case when scale is applied non-uniformly in directions not aligned with local space.
```suggestion
            .svd_unordered(false, false)
            .singular_values
            .max();
```

rasmusgo: The worst case is when we have an extreme non-uniform scaling along the space diagonal. This gives a bound on the error of $\sqrt{3} \approx 1.732 $. We could use this as a fudge factor to guarantee correctness.

kanerogers (approved):
Fantastic work as always @rasmusgo

PR #297, opened 2022-07-29
Add `Stage` component to allow manipulating the transform of the player's frame of reference

### LINKED ISSUE #295: [Feature] Track player's frame of reference as an entity in World
If we make the transform hierarchy for the player look like:
```
               ,-> HMD
World --> Room --> Left Controller
               `-> Right Controller
```
then many kinds of locomotion for the player become a lot easier to implement; you just transform the `Room` (Stage in openxr parlance?) in the world.

Implementation requires synchronising the HMD's transform to the world, and back again in order to get the correct views.

ickk:
Makes the implementation of all kinds of locomotion a lot simpler - just
transform the `Room` in space. 

Should work for teleportation as well as more granular kinds.

Includes systems for correcting the views (`update_views_with_room_transform`) and for adding the Room entity & parenting Hands (`add_room`).

Closes #295


ickk:
open to alternative names for junk

Ralith:
I suggest naming this `Stage` rather than `Room` for consistency with [OpenXR terminology](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#spaces0), unless the intention is to represent something else.

kanerogers:
Agreed with @Ralith here

kanerogers:
@ickk thanks for putting this one through!

I'm trying to wrap my head around this. Why would we represent this as `Stage` instead of something like `Player`? It makes much more sense to say "the view is translated with respect to the `Player`" or "the hands are translated with respect to the `Player`" than it is to say "translated with respect to the `Stage`. It's also just difficult to wrap my head around the idea of a "stage" - think of a theatre or something - that is usually fixed, moving through space.

I've also left a couple nitpicks in the code.

hotham/src/systems/stage.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+    cmd_buffer.run_on(world);
+
+    stage_e
+}
+
+/// Update player's views to take into account the current position of the Stage in the world
+///
+/// Must happen each tick after parent transforms have been updated.
+pub fn update_views_with_stage_transform(world: &mut World, views: &mut [xr::View]) {
```
kanerogers, line 28: This feels super gross (going from `TransformMatrix` to `Isometry3`) but I suppose it's kind of unavoidable until #291 ships. ☹️

kanerogers (changes requested):
@ickk thanks for putting this one through!

I'm trying to wrap my head around this. Why would we represent this as `Stage` instead of something like `Player`? It makes much more sense to say "the view is translated with respect to the `Player`" or "the hands are translated with respect to the `Player`" than it is to say "translated with respect to the `Stage`. 

It's also just difficult to wrap my head around the idea of a "stage" - think of a theatre or something - that is usually fixed, moving through space.

ickk:
> Why would we represent this as Stage instead of something like Player?

It's a little weird to think about, but the player kind of moves independently within the stage, since the thing that most represents the player is probably the HMD, or some IK model derived from the HMD & controllers. However the controllers and HMD are tracked totally independently to each other, but they are tracked with respect to the stage - which makes this approach is much simpler.

Additionally, other kinds of locomotion, such as teleportation, often only really make sense when you teleport the whole stage - especially room-oriented experiences like Owlchemy's stuff.

If it makes you feel better, it's equivalent to moving the world with respect to the chaperone bounds, like how a theatre stage transports to different places in some theatrical world.

hotham/src/systems/stage.rs
ickk: it is mostly just lazy/simpler to reason about imo. We don't need any of the extra abilities that a matrix gives you, and we ultimately have to cast to a posef anyway, which basically is an isometry

rasmusgo: I think it would make sense to let `xr:View` keep `stage_from_view` and add something else that keeps track of `world_from_view`.

ickk: My understanding of how this works is openxr hands us a new view each tick (ideally right before we're about to render), then we either change the views using this system.. or we don't.. Then we just pass them directly to the renderer

I'm happy to have this take a `&xr::View` instead of a `&mut` and return a new owned `xr::View` so you can have a view with respect to both World and Stage, but I'm curious what's the usecase?

kanerogers:
> > Why would we represent this as Stage instead of something like Player?
> 
> It's a little weird to think about, but the player kind of moves independently within the stage, since the thing that most represents the player is probably the HMD, or some IK model derived from the HMD & controllers. However the controllers and HMD are tracked totally independently to each other, but they are tracked with respect to the stage - which makes this approach is much simpler.

This makes my head hurt, but you're right.

> If it makes you feel better, it's equivalent to moving the world with respect to the chaperone bounds, like how a theatre stage transports to different places in some theatrical world.

I appreciate your gentle pandering; that is a wonderful analogy!

kanerogers (approved):
This is a big improvement, but a mental adjustment for my small brain. Nice work!

rasmusgo:
The audio spatialization needs to be updated to work together with this. [audio_system](https://github.com/leetvr/hotham/blob/main/hotham/src/systems/audio.rs#L15) gets the view from `xr_context` with velocity and assumes that the stage and global coordinate systems are the same.

ickk:
oh that's pretty important 😅

Issue #167, opened 2022-03-03
[Rendering] Switch to infinite perspective perspective projection


kanerogers:
## Description
> it seems like the obviously correct default to me
- @Ralith

> I just don't want to see any damn far planes in games any more
- @Ralith

## TODO
- [x] Find correct matrix (should be in HelloOpenXR)
- [ ] Implement

kanerogers:
Projection Matrix, courtesy of @Ralith:

```rust
#[derive(Debug, Copy, Clone)]
pub struct Frustum {
    pub left: f32,
    pub right: f32,
    pub up: f32,
    pub down: f32,
}

impl Frustum {
    #[rustfmt::skip]
    /// Compute right-handed y-up inverse Z perspective projection matrix
    pub fn projection(&self, znear: f32) -> na::Projective3<f32> {
        // Based on http://dev.theomader.com/depth-precision/ + OpenVR docs
        let left = self.left.tan();
        let right = self.right.tan();
        let down = self.down.tan();
        let up = self.up.tan();
        let idx = 1.0 / (right - left);
        let idy = 1.0 / (down - up);
        let sx = right + left;
        let sy = down + up;
        na::Projective3::from_matrix_unchecked(
            na::Matrix4::new(
                2.0 * idx, 0.0, sx * idx, 0.0,
                0.0, 2.0 * idy, sy * idy, 0.0,
                0.0,       0.0,      0.0, znear,
                0.0,       0.0,     -1.0, 0.0))
    }
}

impl From<xr::Fovf> for Frustum {
    fn from(x: xr::Fovf) -> Self {
        Self {
            left: x.angle_left,
            right: x.angle_right,
            up: x.angle_up,
            down: x.angle_down,
        }
    }
}
```

PR #298, opened 2022-07-30
Employ a reversed infinite perspective projection

Ralith:
Avoids clipping of distant objects and minimizes Z-fighting. Also, bonus cleanup.

Validated by running crab-saber, which looks fine.

Fixes https://github.com/leetvr/hotham/issues/167


kanerogers (approved):
LGTM!

PR #299, opened 2022-07-31
Fix tests

rasmusgo:
Some tests are not run by CI and only work on windows so they have gotten broken during development. I have fixed all ~~but `test_audio_system`~~ in this PR.

`test_audio_system` had some openxr lifecycle issue. I just pattern matched myself to something that made the test pass. @kanerogers, it would be great if you looked at that extra carefully since you worked on the openxr lifecycle recently.


kanerogers:
Fixed a minor issue with `test_draw_gui`

kanerogers (approved):
Looks good to me 👍🏻

PR #305, opened 2022-08-03
Simulator: Fix accumulation of mouse movement

rasmusgo:
* Remove influence of `dt `on mouse sensitivity.
* Accumulate mouse events instead of overwriting.


kanerogers (approved):
Thanks! I noticed some pretty insane mouse movement stuff.

PR #306, opened 2022-08-03
Fix clip planes

### LINKED ISSUE #303: [Rendering] Culling does not appear to be correct
## Background
In this screenshot, the camera is facing away from the asteroid:

image

However, its draw call can be seen in RenderDoc:

image

_But_, `EID` 37 shows that about half the cubes have been culled (eg. the draw command is `vkCmdDrawIndexed(36, 655)`). Given that there are a total of 1,653 cubes, it would follow that these cubes have been culled correctly.

rasmusgo:
This resolves #303. @kanerogers please verify!


kanerogers (approved):
looks good

PR #308, opened 2022-08-04
Fix Hands-RigidBody desync

### LINKED ISSUE #300: [Feature] `Stage` as entity follow up work
- ~~As pointed out by @rasmusgo, our audio system needs to take into account the position of the Stage in the world~~
moved to #309

- [x] Currently the rigid bodies & transforms of the hands are desynchronised when the Stage moves away from the origin. This needs to be addressed

ickk:
Closes #300

This removes the hierarchy from the Stage, instead just manually offset the positions of the hands by the stage. (At this point we have elected to disallow using both physics and ecs-hierarchies simultaneously)

Also adds a rigid body to the stage, to make everything consistent. This prevents oddities in apps where you may place a system, which depends on both Hands & Stage, between the transform update systems and the physics system.

this points at a PR on hotham

### DIFF
--- a/hotham/src/systems/hands.rs
+++ b/hotham/src/systems/hands.rs
@@ -1,10 +1,10 @@
 use crate::{
     asset_importer::add_model_to_world,
-    components::{hand::Handedness, AnimationController, Hand, RigidBody},
+    components::{hand::Handedness, AnimationController, Hand, RigidBody, Stage},
     resources::{PhysicsContext, XrContext},
     util::{is_space_valid, posef_to_isometry},
 };
-use hecs::{PreparedQuery, World};
+use hecs::{PreparedQuery, With, World};
 use rapier3d::prelude::{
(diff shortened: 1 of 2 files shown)


hotham/src/systems/hands.rs
```diff
@@ -43,7 +50,10 @@ pub fn hands_system(
             .unwrap();
 
         let position = posef_to_isometry(pose);
-        rigid_body.set_next_kinematic_position(position);
+        match stage_isometry {
+            Some(stage) => rigid_body.set_next_kinematic_position(stage * position),
+            None => rigid_body.set_next_kinematic_position(position),
+        }
```
rasmusgo, line 56: Consider using `y_from_x` for the transforms, eg:
```rs
        let stage_from_local = posef_to_isometry(pose);
        let global_from_stage = match stage_isometry {
            Some(stage) => stage,
            None => identity,
        };
        rigid_body.set_next_kinematic_position(global_from_stage * stage_from_local),
```

kanerogers (approved):
Nice work!

PR #311, opened 2022-08-05
Preventing keyboard inputs when window isn't focused

louisgjohnson:
Fix: This is to address a bug in the simulator where winit is listening to keyboard inputs when window isn't focused.

I've changed the winit event listening from device events to window events for keyboard inputs

I also refactored the camera update functions to split them into separate functions for rotation and position updates..


kanerogers (approved):
Nice one, @louisgjohnson ! 👍🏻 

nitpick: Good to always have a reference issue before working on a PR, and best to keep PRs focussed. Cleanups can always be done in a separate PR so they can be reviewed separately 😄

PR #312, opened 2022-08-07
Holographic materials

rasmusgo:
This is so exciting that I can hardly contain myself!

What I mean with holographic materials is that they contain a 3D image even when rendered using a single flat triangle. This embedded image is raytraced by the fragment shader when the triangle is rasterized. This PR introduces holographic spheres but I look forward to adding more shapes such as cylinders and capsules.

I have added a sphere to the simple scene example:

image

The polygonal hull that is used for rendering this hologram is only 20 triangles:

image

There is some more work to be done with UV-mapping etc. but I think this is a good proof of concept!


rasmusgo:
This will be remade in a more efficient way as described in #326.

PR #313, opened 2022-08-07
Fix asset_importer::tests::test_hand

rasmusgo:
I thought I fixed this before but it seems to have gotten lost.


kanerogers (approved):
Yes, I remember this being changed as well 😕 

Best not to think about it too much..

Issue #317, opened 2022-08-09
[Simulator] Virtual input support


kanerogers:
## Background
Now that we've got input handled better inside Hotham, let's add something to the simulator.

First pass is to press the X, Y, A and B buttons.

## TODO
- [x] Push button
- [x] Receive input

## Follow up tasks
- [ ] #318

PR #319, opened 2022-08-09
Kanerogers/issue317

kanerogers:
## What is this related to?
#317

## What changed?
- The simulator now maps keyboard input to OpenXR actions
- This first implementation is limited to the X/Y/A/B buttons on the touch controller, but future work can go further.

- Developers can now test some input interactions with the simulator

### DIFF
--- a/hotham-simulator/src/action_state.rs
+++ b/hotham-simulator/src/action_state.rs
@@ -1,0 +1,34 @@
+use std::collections::HashMap;
+
+use openxr_sys::{Action, Path, FALSE};
+
+#[derive(Debug, Clone, Default)]
+/// Stores Action state, allowing the simulator to simulate input for an application.
+// A bit yuck to use u64 instead of Action, but it doesn't support Hash.. but whatever.
+pub struct ActionState {
+    boolean_actions: HashMap<u64, bool>,
+    bindings: HashMap<Path, u64>,
+}
+impl ActionState {
+    pub(crate) fn get_boolean(&self, action: Action) -> openxr_sys::Bool32 {
+        self.boolean_actions
+            .get(&action.into_raw())
+            .map(|p| (*p).into())
+            .unwrap_or(FALSE)
+    }
+
+    pub(crate) fn add_binding(&mut self, path: Path, action: Action) {
+        self.bindings.insert(path, action.into_raw());
+    }
+
+    /// Resets all action state.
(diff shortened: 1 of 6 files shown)


hotham-simulator/src/action_state.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use openxr_sys::{Action, Path, FALSE};
+
+#[derive(Debug, Clone, Default)]
+/// Stores Action state, allowing the simulator to simulate input for an application.
+// A bit yuck to use u64 instead of Action, but it doesn't support Hash.. but whatever.
+pub struct ActionState {
+    boolean_actions: HashMap<u64, bool>,
+    bindings: HashMap<Path, u64>,
+}
```
rasmusgo, line 11: I think we should be able to [use the newtype pattern](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types) to implement `Hash` for `Action`.

hotham-simulator/src/state.rs
```diff
@@ -20,8 +20,15 @@ use std::{
...
-use crate::{inputs::Inputs, simulator::NUM_VIEWS, space_state::SpaceState};
-// use crate::simulator::spa
+use crate::{
+    action_state::ActionState, inputs::Inputs, simulator::NUM_VIEWS, space_state::SpaceState,
+};
+
+static X_INPUT: &str = "/user/hand/left/input/x/click";
+static Y_INPUT: &str = "/user/hand/left/input/y/click";
+static B_INPUT: &str = "/user/hand/right/input/b/click";
+static A_INPUT: &str = "/user/hand/right/input/a/click";
```
rasmusgo, line 30: I think we should stick with alphabetic order as much as possible to keep things consistent.

hotham-simulator/src/action_state.rs
kanerogers: We could, and that was my first thought, but I think that's unnecessary. We've got type safety anyway in the public API of `ActionState`, I don't think there's much to be gained otherwise.

hotham-simulator/src/state.rs
kanerogers: Ah that's reasonable!

rasmusgo (approved):

PR #321, opened 2022-08-09
Refactor vulkan buffer bindings

rasmusgo:
Tidy up Vulkan buffer bindings a bit by giving them names instead of using magic numbers.

### DIFF
--- a/hotham/src/rendering/buffer.rs
+++ b/hotham/src/rendering/buffer.rs
@@ -131,7 +131,7 @@ impl<T: Sized> Buffer<T> {
         &self,
         device: &ash::Device,
         descriptor_set: vk::DescriptorSet,
-        binding: usize,
+        binding: u32,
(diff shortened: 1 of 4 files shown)


hotham/src/rendering/buffer.rs
```diff
@@ -131,7 +131,7 @@ impl<T: Sized> Buffer<T> {
         &self,
         device: &ash::Device,
         descriptor_set: vk::DescriptorSet,
-        binding: usize,
+        binding: u32,
```
kanerogers, line 134: Oh good pickup! 🤦🏻

kanerogers (approved):
Nice one @rasmusgo~! I had originally been a bit cautious about this as the magic numbers really do have value; they have specific representations in the shaders.

But I think this makes much more sense. Thank you!

PR #322, opened 2022-08-09
Use dummy keystore for examples

rasmusgo:
This makes it easier to get started with hotham development because the user doesn't need to create a keystore to get the examples to run.

You may need to uninstall the examples from your quest if you have installed them before with a diffrent keystore.


kanerogers:
Nice one! My only quibble is the warning message in the `Cargo.toml`. I think I'd prefer something like:

```toml
# !! IMPORTANT !! 
#
# When creating your own apps, make sure to generate your own keystore, rather than using our example one!
# You can use `keytool` like so:
# keytool -genkey -v -keystore my-release-key.keystore -keyalg RSA -keysize 2048 -validity 10000
# 
# For more information on key signing and why it's so important, check out this article: 
# https://developer.android.com/studio/publish/app-signing
#
# !! IMPORTANT !! 
[package.metadata.android.signing.release]

```

kanerogers (changes requested):
Nice one! My only quibble is the warning message in the `Cargo.toml`. I think I'd prefer something like:

```toml
# !! IMPORTANT !! 
#
# When creating your own apps, make sure to generate your own keystore, rather than using our example one!
# You can use `keytool` like so:
# keytool -genkey -v -keystore my-release-key.keystore -keyalg RSA -keysize 2048 -validity 10000
# 
# For more information on key signing and why it's so important, check out this article: 
# https://developer.android.com/studio/publish/app-signing
#
# !! IMPORTANT !! 
[package.metadata.android.signing.release]

```

kanerogers (approved):
LGTM!

Issue #255, opened 2022-07-11
[Rendering] Add IBL back


kanerogers:
## Background
Now that we've got some more texture bandwidth to play with, let's add some sweet looking IBLs.

## TODO
- [x] Add back fragment shader code
- [x] Wire up descriptors
- [x] Load static IBL assets

## Further reading
- https://github.com/KhronosGroup/3DC-Certification/tree/main/documents

kanerogers:
Current update: Loading cubemaps is annoying.

kanerogers:
Bad:

![image](https://user-images.githubusercontent.com/2022375/183657309-6809cf41-39c7-42b3-b8ad-c3b522eda8bd.png)

Good (but technically wrong):

![Uploading image.png…]()

Issue #314, opened 2022-08-07
[Rendering] Implement support for glTF lighting


kanerogers:
## Background
glTF implements the usual suspects of realtime dynamic lighting. Let's allow users to import them into their scenes.

## References
- [KHR_lights_punctual](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_lights_punctual/README.md)
- [Khronos glTF Sample Viewer](https://github.com/KhronosGroup/glTF-Sample-Viewer/blob/78e6453306923f1c0df3220d45a2e0656b80c326/source/Renderer/shaders/pbr.frag#L179)
- [gltf-viewer-rs](https://github.com/adrien-ben/gltf-viewer-rs/blob/master/crates/viewer/shaders/model.frag)

## TODO
- [x] Implement a hardcoded point light
- [x] Implement a hardcoded spotlight
- [x] #324 
- [x] Add 4 point light members to `SceneParams`
- [x] Fix normals on hands
- [x] Fix IBL

kanerogers:
Hardcoded lights working well; after some refactoring to separate analytical light from IBL, adding the different intensity functions for various lights is pretty trivial. Next step is to add an interface to add lights from the CPU, then after that we can extract them from a glTF file.

kanerogers:
It looks like we're definitely not handling normals correctly. This is how the Normal Tangent Mirror test should look with a single point light in the middle of the scene:

![Normals Good](https://user-images.githubusercontent.com/2022375/184250612-3a52bbc4-47e4-4bfb-b9b0-081528dbd8eb.png)

..and this is how it looks in Hotham:
![Normals Bad](https://user-images.githubusercontent.com/2022375/184250711-f89158e6-bc8a-434c-be0c-a5ccd5c75a0d.png)

kanerogers:
Closed in #327

PR #323, opened 2022-08-10
Add back Image Based Lighting.

kanerogers:
## What is this related to?
#255

## What changed?
- Added back support for Image Based Lighting
- Scenes now have significantly improved visual fidelity
- Added a `debug_system` to debug shader inputs
- Modified `SceneParams` to change `debug_data` to `params`, allowing users to modify the intensity of IBL
- Added a Khronos supplied scene to test IBL

- Scenes should look a lot better
- There will be a fairly significant performance impact, but optimisations can be made

### Before
image

### After
image


rasmusgo:
Is it just me or does this look too washed out with IBL?

hotham/src/rendering/scene_data.rs
```diff
@@ -11,26 +11,24 @@ pub struct SceneData {
...
 }
 
 /// Create a new directional light
-pub fn new_directional_light(x: f32, y: f32) -> Vector4<f32> {
-    let x = x.sin() * y.cos();
-    let y = y.sin();
-    let z = x.cos() * y.cos();
-    [x, y, z, 1.].into()
+pub fn new_directional_light(x: f32, y: f32, z: f32) -> Vector4<f32> {
+    [x.to_radians(), y.to_radians(), z.to_radians(), 1.].into()
```
rasmusgo, line 33: This is very counter intuitive. The parameter names implies a postion but we are treating them as some kind of angles.

hotham/src/shaders/pbr.frag
```diff
@@ -84,33 +92,28 @@ vec3 getNormal(uint normalTextureID)
...
-// 	// return diffuse + specular;
-// 	return vec3(1.);
-// }
+	// retrieve a scale and bias to F0. See [1], Figure 3
+	vec2 brdfSamplePoint = clamp(vec2(pbrInputs.NdotV, 1.0 - pbrInputs.perceptualRoughness), vec2(0.0, 0.0), vec2(1.0, 1.0));
+	vec3 brdf = texture(textures[BRDF_LUT_TEXTURE_ID], brdfSamplePoint).rgb;
+
+	// Get diffuse and specular light values
+	vec3 diffuseLight = tonemap(texture(cubeTextures[SAMPLER_IRRADIANCE_TEXTURE_ID], n)).rgb;
+	vec3 specularLight = tonemap(textureLod(cubeTextures[ENVIRONMENT_MAP_TEXTURE_ID], reflection, lod)).rgb;
```
rasmusgo, line 109: If you tonemap here, it does not make sense to blend it later on because tonemapping is non-linear.

hotham/src/shaders/pbr.frag
```diff
@@ -286,27 +267,27 @@ void main()
...
 	vec3 diffuseContrib = (1.0 - F) * diffuse(pbrInputs);
 	vec3 specContrib = F * G * D / (4.0 * NdotL * NdotV);
+
 	// Obtain final intensity as reflectance (BRDF) scaled by the energy of the light (cosine law)
 	vec3 color = NdotL * (diffuseContrib + specContrib);
 
-	// Calculate lighting contribution from image based lighting source (IBL)
-	color += getIBLContribution(pbrInputs, n, reflection) * DEFAULT_IBL_SCALE;
+	// Calculate lighting contribution from image based lighting source (IBL), scaled by a scene data parameter.
+	color += getIBLContribution(pbrInputs, n, reflection) * sceneData.params.x;
```
rasmusgo, line 275: Could be nice to avoid calling this function if `sceneData.params.x` is zero. Not sure how good that works in a shader though.

hotham/src/rendering/scene_data.rs
rasmusgo: I think the `.to_radians()` should be removed and the vector should end with zero instead of one because it is a direction and not a position. Ending with zero makes it behave the correct way when transformed.

rasmusgo (changes requested):

rasmusgo:
There is a colorchecker board showing in the reflections in the IBL test scene. It shows as shades of gray though. Is this expected? Is the environment map a grayscale image?

image

hotham/src/rendering/scene_data.rs
kanerogers: Ah, totally ignore this, I'm rewriting all the light functions in #314. This has always been broken 😆

hotham/src/shaders/pbr.frag
kanerogers: Yep, agreed

hotham/src/shaders/pbr.frag
kanerogers: Yep, and this is fixed in #314

kanerogers:
> There is a colorchecker board showing in the reflections in the IBL test scene. It shows as shades of gray though. Is this expected? Is the environment map a grayscale image?
> 
> image

Yeah, this environment map is greyscale.

rasmusgo (approved):

PR #325, opened 2022-08-10
Ibl suggestions

rasmusgo:
Here are some suggestions for the image based lighting PR. Feel free to cherry pick what you like or merge this into #323 if you like all of it.


kanerogers:
Thanks so much for putting this together! I've broadly incorporated these changes in #314 !

PR #330, opened 2022-08-14
Rasmusgo/refactor pbr shader

rasmusgo:
This makes it easier to make new shaders that use the same PBR shading.

### DIFF
--- a/hotham/src/rendering/resources.rs
+++ b/hotham/src/rendering/resources.rs
@@ -202,8 +202,9 @@ fn load_ibl_textures(
 pub struct DrawData {
     /// The transform of the parent mesh
     pub global_from_local: Matrix4<f32>,
-    /// The inverse transpose of the transform of the parent mesh
-    pub inverse_transpose: Matrix4<f32>,
+    /// The inverse of the transform of the parent mesh
+    /// Transform normals by multiplying with the matrix on the right hand side
+    pub local_from_global: Matrix4<f32>,
     /// The ID of the material to use.
     pub material_id: u32,
     /// An optional skin to use.
--- a/hotham/src/systems/rendering.rs
+++ b/hotham/src/systems/rendering.rs
@@ -186,11 +186,7 @@ pub unsafe fn draw_world(vulkan_context: &VulkanContext, render_context: &mut Re
             let instance = &instanced_primitive.instances[cull_result.index_instance as usize];
             let draw_data = DrawData {
                 global_from_local: instance.global_from_local,
-                inverse_transpose: instance
-                    .global_from_local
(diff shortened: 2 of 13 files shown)


hotham/src/rendering/resources.rs
```diff
@@ -203,7 +203,7 @@ pub struct DrawData {
     /// The transform of the parent mesh
     pub global_from_local: Matrix4<f32>,
     /// The inverse transpose of the transform of the parent mesh
-    pub inverse_transpose: Matrix4<f32>,
+    pub local_from_global: Matrix4<f32>,
```
kanerogers, line 206: But is this actually `local_from_global`? It's also transposed

hotham/src/systems/rendering.rs
```diff
@@ -154,11 +154,7 @@ pub fn rendering_system(
                 let instance = &instanced_primitive.instances[cull_result.index_instance as usize];
                 let draw_data = DrawData {
                     global_from_local: instance.global_from_local,
-                    inverse_transpose: instance
-                        .global_from_local
-                        .try_inverse()
-                        .unwrap()
-                        .transpose(),
+                    local_from_global: instance.global_from_local.try_inverse().unwrap(),
```
kanerogers, line 157: I don't think this is correct! Everything I've read about these normals indicates they need a transpose. 😨

rasmusgo: I'm applying it from the left, treating the normal as a row vector. See here: https://github.com/leetvr/hotham/pull/330/files#diff-02130a77bf74f1b203443bd19cca29f2adcc1dd8d7b991bc2b83f39d1589c976R32

rasmusgo: This is arguably the more correct way of treating normals because they are bivectors. Mathematical bike shedding perhaps 😛

kanerogers: So as a mathematical simpleton, why do other implementations use the inverse transpose, when what you're suggesting is far simpler?

rasmusgo: Maybe because it's hard to make math simple? Bivectors are not commonly taught and this can be solved without them. Few people are curious enough to keep looking for better ways to express math once they have something that works.

rasmusgo:
Will reverting  `Pass p, n, uv...` [1032ba2](https://github.com/leetvr/hotham/pull/330/commits/1032ba22de6f588121bef1768e77eb3c77b66d67) get this unstuck?

kanerogers (approved):
LGTM! Nice work as always @rasmusgo 🎉

PR #336, opened 2022-08-16
Use stage space instead of global space when rendering

rasmusgo:
Using stage as the main coordinate system during rendering keeps numbers in the same range even when moving around big worlds. This should help to avoid problems with numerical precision especially if we start to use floats with fewer bits in the shader

> Yes. This is an important optimisation that we need to implement sooner rather than later. 32bit operations require two instructions on the Adreno 650
> I'm no mathematician, but if we can go from two instructions to one that means that we can save our instructions by.
> ..
> uh..
> 50%!

 — Kane


hotham/src/systems/stage.rs
```diff
@@ -23,20 +21,3 @@ pub fn add_stage(world: &mut World, physics_context: &mut PhysicsContext) -> Ent
         rigid_body,
     ))
 }
-
-/// Update player's views to take into account the current position of the Stage in global space
-///
-/// Must happen each tick after parent transforms have been updated.
-pub fn update_views_with_stage_transform(world: &mut World, views: &mut [xr::View]) {
```
rasmusgo, line 30: This function should not be called anymore because the views are already in the correct coordinate system.

kanerogers:
Let me check this with The Station and get back to you!

rasmusgo:
The image based lighting does not play well with this because it relies on the coordinates being aligned with the axes of global space. We could compensate for this in the fragment shader but at that point the added complexity might be better spent rendering in globally oriented stage space (GOS).

rasmusgo:
Superseded by #355.

PR #338, opened 2022-08-16
Fix light attenuation

### LINKED ISSUE #335: [Rendering] Fix spotlight attentuation
## Background
If you have some spotlights 400m away, the scene should look like this:
Image

Not like this:
Image

rasmusgo:
The math is a bit tricky in this one 😉 

Surprisingly, this did not trip the pbr test. You will most likely need to tune your light intensities though.

This resolves #335


kanerogers (approved):
Good pickup! 😀 Thank you very much!

re: PBR test, that's not surprising as it doesn't have any punctual lights, just IBL. Might be good to add some lights to the test, though!

Issue #294, opened 2022-07-28
[Bug] Updating transform of skinned joints on quest 2 is broken.. somehow


ickk:
There is validation that skinning *appears* to work on [ickk/skinning-sanity-test](https://github.com/leetvr/hotham/tree/ickk/skinning-sanity-test), however there in certain circumstances this seems to fail. 

Everything seems to work fine in the simulator, but it fails on device.

This could be related to #234.

Reproduction is available on https://github.com/leetvr/the_station/issues/47

Exact cause remains elusive

kanerogers:
@ickk I have a strange feeling this may be related to re-ordering of systems. Will investigate

kanerogers:
Fixed in #340

PR #340, opened 2022-08-23
Fix hands animations on quest

### LINKED ISSUE #234: [Bug] Hands animation no longer working
## Background
After the upgrade of the glTF importer and the new rendering backend, the hands animation is no longer working. This is somewhat understandable as the way hands were being animated was a big hack.

glTF animation is [well defined](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#animations), but for some reason when exporting the default Oculus hands from Unity, the animations are completely mangled: instead of giving us _keyframes_ between two states (hand open and hand closed) we get two _animations_, each with a **single** keyframe. We work around this by just creating a single animation and use each state (open and closed) as a keyframe.

## TODO
- [ ] Should we just get better models with correct animations?

rasmusgo:
This resolves #234 and #294.

I'm guessing this is a driver/compiler bug...

### DIFF
--- a/hotham/src/shaders/pbr.vert
+++ b/hotham/src/shaders/pbr.vert
@@ -35,12 +35,11 @@ void main() {
         outTBN = mat3(globalTangent, globalBiTangent, globalNormal);
     } else {
         // Mesh is skinned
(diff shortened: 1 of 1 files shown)


kanerogers (approved):
hero

PR #342, opened 2022-08-27
Compute tangents inside shader again

rasmusgo:
Computing tangents on the fly instead of passing them along in the vertex data saves memory bandwidth and is actually faster. This also gets rid of a newly added dependency.

Helmets stress test:
image
[helmets-main.log](https://github.com/leetvr/hotham/files/9438667/helmets-main.log)
[helmets-tangents-from-derivatives.log](https://github.com/leetvr/hotham/files/9438668/helmets-tangents-from-derivatives.log)

Vertices stress test:
image
[vertices-main.log](https://github.com/leetvr/hotham/files/9438669/vertices-main.log)
[vertices-tangents-from-derivatives.log](https://github.com/leetvr/hotham/files/9438670/vertices-tangents-from-derivatives.log)

I have looked at the `NormalTangentTest` and it looks correct from what I can tell.

`test_rendering_pbr` produces similar normals but there are some slight differences on the small features close to the edge of the helmet
render_Normals


hotham/src/shaders/pbr.glsl
```diff
@@ -56,22 +56,25 @@ vec3 tonemap(vec3 color) {
...
+        return N;
     }
+
+    vec3 textureNormal;
+    textureNormal.xy = texture(textures[normalTextureID], inUV).ga * 2.0 - 1.0;
+    textureNormal.z = sqrt(1 - dot(textureNormal.xy, textureNormal.xy));
+
+    vec3 dGlobalPosDx = dFdx(inGlobalPos);
+    vec3 dGlobalPosDy = dFdy(inGlobalPos);
+    vec2 dUvDx = dFdx(inUV);
```
kanerogers, line 70: nit: indentation seems off

hotham/src/shaders/pbr.glsl
```diff
@@ -56,22 +56,25 @@ vec3 tonemap(vec3 color) {
...
-        ntex.z = sqrt(1 - dot(ntex.xy, ntex.xy));
-        return normalize(mat3(t, b, ng) * ntex);
-    } else {
-        return ng;
+    vec3 N = normalize(inNormal);
+    if (normalTextureID == NOT_PRESENT) {
+        return N;
     }
+
+    vec3 textureNormal;
```
kanerogers, line 64: Could you add a couple comments here? Just something like "compute the tangent based on..". May also be worth noting the performance benefit in case someone looks at this and thinks it may be slower

kanerogers (changes requested):
@rasmusgo once again you have proven me wrong 😛 Fantastic work, and really appreciate the performance deltas - can't argue with numbers! This certainly proves your point that vertex fetching is expensive and should be minimised: I wonder what else we could strip away from the vertex?

kanerogers (approved):
LGTM!!

PR #343, opened 2022-08-30
Rasmusgo/wait for pid in scripts

rasmusgo:
* Wait for app to start before running `logcat` in bash scripts.
* Add bash script for running stress test.

Tested on windows with git bash.


kanerogers (approved):
LGTM! Nice thinking. I'll whip together a powershell version.. someday.

PR #344, opened 2022-09-01
Fix saber-grip offset

rasmusgo:
Have you noticed that the sabers move weirdly when you rotate the controllers? Well, I have and this PR fixes it by applying the rotation and translation in the correct order. I also threw in a stage transformation just for completeness.


kanerogers (approved):
Nice! Grey pickup!

PR #348, opened 2022-09-10
Profiling with puffin

rasmusgo:
image

This PR adds [puffin ](https://github.com/EmbarkStudios/puffin) and [puffin_http](https://github.com/EmbarkStudios/puffin/tree/main/puffin_http) for profiling. This allows for remote profiling and viewing profiling data live with [puffin_viewer](https://github.com/EmbarkStudios/puffin/tree/main/puffin_viewer). This works both locally when running in desktop mode and remotely when running natively on the quest. The viewer is currently always on the desktop side but an in-game viewer ([puffin_egui](https://github.com/EmbarkStudios/puffin/tree/main/puffin_egui)) could be added later on.

Install and run the remote viewer with:
```sh
cargo install puffin_viewer
puffin_viewer --url 127.0.0.1:8585
```

Then run an application with profiling enabled. The stress-test enables profiling automatically and it is a single function call to enable profiling in any other application.

You will need to find the IP address of the quest to connect to it when profiling natively. I found it by looking at the device list in my router but there are probably better ways to do this.

### DIFF
--- a/benchmarks/stress-test/src/lib.rs
+++ b/benchmarks/stress-test/src/lib.rs
@@ -43,6 +43,8 @@ pub fn real_main() -> HothamResult<()> {
     let queries = Default::default();
     let timer = Default::default();
 
+    engine.start_puffin_server();
+
     let mut tick_props = TickProps {
         engine,
         world,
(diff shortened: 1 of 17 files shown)


kanerogers:
As per my discord message, while puffin definitely looks interesting, I think we're much better off looking at [tracy](https://github.com/wolfpld/tracy) instead on account of its specific focus on gamedev, and strong Vulkan integration.

PR #349, opened 2022-09-10
Fix logo.png git lfs files

rasmusgo:
`benchmarks/stress-test/logo.png` was committed as a binary file but according to `.gitattributes` it should be a git lfs pointer file. This causes git to be confused and thinking the file is modified locally even when it isn't. This fixes the problem by adding the file as a git lfs pointer file instead. Same for `examples/simple-scene/logo.png`.


kanerogers (approved):
Stupid git lfs. Thank you!

PR #355, opened 2022-09-19
Render in globally oriented stage space

rasmusgo:
This replaces #336.

The Globally Oriented Stage (GOS) space is global space translated so that the origin coincides with the origin of stage space. Using GOS instead of global space keeps the values of vertex positions small for vertices close to the player. This makes it less risky to use half precision floats. The winning argument to use GOS over stage space for rendering was that environment maps (eg. for image based lighting) still works as expected because the axis directions are the same as in global space.

### DIFF
--- a/hotham/src/rendering/camera.rs
+++ b/hotham/src/rendering/camera.rs
@@ -6,7 +6,7 @@ use crate::util::posef_to_isometry;
 #[derive(Debug, Clone)]
 /// The Camera, or View, in a scene.
 pub struct Camera {
-    /// The camera's position in globally oriented stage space
+    /// The camera's pose in globally oriented stage space
     pub gos_from_view: Isometry3<f32>,
     /// The view matrix
     pub view_from_gos: Matrix4<f32>,
@@ -23,9 +23,9 @@ impl Default for Camera {
 }
 
 impl Camera {
-    /// Update the camera's position from an OpenXR view (stage from view)
+    /// Update the camera's position from an OpenXR view
     pub fn update(&mut self, view: &View, gos_from_stage: &Isometry3<f32>) -> Matrix4<f32> {
-        // Convert values from OpenXR format
+        // Convert values from OpenXR format and use globally oriented stage space instead of stage space
         let stage_from_view = posef_to_isometry(view.pose);
         self.gos_from_view = gos_from_stage * stage_from_view;
 
@@ -34,7 +34,7 @@ impl Camera {
     }
 
     /// Get the camera's position
-    pub fn position(&self) -> Vector4<f32> {
+    pub fn position_in_gos(&self) -> Vector4<f32> {
         let p = self.gos_from_view.translation.vector;
         vector![p[0], p[1], p[2], 0.]
     }
--- a/hotham/src/rendering/camera.rs
+++ b/hotham/src/rendering/camera.rs
@@ -6,41 +6,41 @@ use crate::util::posef_to_isometry;
 #[derive(Debug, Clone)]
 /// The Camera, or View, in a scene.
 pub struct Camera {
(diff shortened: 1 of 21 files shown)


hotham/src/rendering/camera.rs
```diff
@@ -6,41 +6,41 @@ use crate::util::posef_to_isometry;
...
+        self.gos_from_view = gos_from_stage * stage_from_view;
 
-        self.view_matrix = self.build_matrix();
-        self.view_matrix
+        self.view_from_gos = self.build_matrix();
+        self.view_from_gos
     }
 
     /// Get the camera's position
     pub fn position(&self) -> Vector4<f32> {
```
kanerogers, line 37: I think that this function should be renamed to something like `gos_position` to keep it consistent with the rest of the engine. All other position related functions and variables in Hotham return global space positions

hotham/src/rendering/camera.rs
```diff
@@ -6,41 +6,41 @@ use crate::util::posef_to_isometry;
 #[derive(Debug, Clone)]
 /// The Camera, or View, in a scene.
 pub struct Camera {
```
kanerogers, line 8: It's still sometimes helpful to know the position of the camera; I'd leave the `position` variable here.

hotham/src/rendering/light.rs
```diff
@@ -30,7 +30,7 @@ pub struct Light {
     pub intensity: f32,
 
     /// The position of the light
-    pub position: Vector3<f32>,
+    pub position: Point3<f32>,
```
kanerogers, line 33: Good call.

hotham/src/rendering/camera.rs
```diff
@@ -6,41 +6,41 @@ use crate::util::posef_to_isometry;
...
+            gos_from_view: Isometry3::from_parts(t, UnitQuaternion::identity()),
+            view_from_gos: Matrix4::identity(),
         }
     }
 }
 
 impl Camera {
-    /// Update the camera's position from an OpenXR view
-    pub fn update(&mut self, view: &View) -> Matrix4<f32> {
+    /// Update the camera's position from an OpenXR view (stage from view)
```
kanerogers, line 26: This comment is a bit unclear; I'm not sure it's necessary to mention the stage here.

hotham/src/rendering/camera.rs
```diff
@@ -6,41 +6,41 @@ use crate::util::posef_to_isometry;
...
+    pub fn update(&mut self, view: &View, gos_from_stage: &Isometry3<f32>) -> Matrix4<f32> {
         // Convert values from OpenXR format
-        let camera_position = posef_to_isometry(view.pose);
-        self.position = camera_position;
+        let stage_from_view = posef_to_isometry(view.pose);
```
kanerogers, line 29: It would be good to add some comments explaining what's going on here, as the existing comment mentions format conversions, but there's significant work around spaces here that's not fully explained.

kanerogers (changes requested):
Looking good! I ran some of the stress tests and it seems to be working nicely.

I did make a couple of style / consistency comments that would be good to get addressed before we move on. 😄

hotham/src/rendering/camera.rs
rasmusgo: The position variable was actually a transform with a bad name. I just gave it a better name.

kanerogers: It wasn't *technically* a transform since the camera has no scale. :P But in any case, it would be good to still know where the camera is located, in world space!

rasmusgo: A transform without scale is still a transform.

kanerogers (approved):
LGTM!

Issue #359, opened 2022-09-20
[Maintenance] Use `document.nodes()` where possible in asset importer


kanerogers:
## Background
In most cases, it's not necessary to traverse the whole scene hierarchy to import nodes. In these cases, it's possible to simply use the `document.nodes()` accessor instead.

## TODO
- [ ] Fix `load_skins` to take `document.nodes()`

PR #360, opened 2022-09-20
[WIP] using document.nodes instead of searching through document scenes

louisgjohnson:
Closes: #359 

Using document.nodes to access nodes instead of traversing scene.

### DIFF
--- a/hotham/src/asset_importer/mod.rs
+++ b/hotham/src/asset_importer/mod.rs
@@ -130,7 +130,7 @@ fn load_models_from_gltf_data(import_context: &mut ImportContext) -> Result<()> 
     // We need *some* entity to stash the AnimationController onto.
     // For now, just use the first root entity.
     let mut animation_controller_entity = None;
-    for node in document.nodes() {
+    for node in document.scenes().next().unwrap().nodes() {
         let mut world = World::default();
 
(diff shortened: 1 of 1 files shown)


louisgjohnson:
Found an issue with one of the examples, marked as WIP.

hotham/src/asset_importer/mod.rs
```diff
@@ -134,7 +130,7 @@ fn load_models_from_gltf_data(import_context: &mut ImportContext) -> Result<()>
     // We need *some* entity to stash the AnimationController onto.
     // For now, just use the first root entity.
     let mut animation_controller_entity = None;
-    for node in document.scenes().next().unwrap().nodes() {
+    for node in document.nodes() {
```
kanerogers, line 133: Ah! With this one you definitely need to use `scene.nodes()`. The reason is because `document.nodes()` doesn't have any hierarchy information, so the nodes won't get loaded in correctly. It should be the *one* exception to this though.

louisgjohnson: ah yep that makes sense! well I've reversed it back and tested it with crab_saber and simple_scene and it seems to work fine

kanerogers (approved):
Nice work! LGTM!

PR #362, opened 2022-09-23
Import colliders from glTF

### LINKED ISSUE #351: [Feature] Easy importing of colliders from glTF files
## Background
One of the most time consuming tasks when working on Hotham games is configuring collider geometry. This issue is a first past attempt at resolving that problem.

## TODO
- [ ] When importing meshes, check for `_HOTHAM_COLLIDER_WALL` or `_HOTHAM_COLLIDER_SENSOR` suffixes
- [ ] For `WALL` meshes, create a collider and attach it to the corresponding entity
- [ ] For `SENSOR` meshes, create a collider and mark it as a sensor

kanerogers:
## What is this related to?
Closes #351

## What changed?
- Added the ability to define colliders using mesh geometry, then add a corresponding `Collider` component to the entity
- Initial support is for two kinds of colliders - walls and sensors
- Sensors have no corresponding mesh and are configured as Rapier sensors. These are tagged with the `".HOTHAM_COLLIDER_SENSOR"` suffiix in the node's name.
- Walls are attached as a collider to an entity that has a mesh by adding the `".HOTHAM_COLLIDER_WALL"` suffix to the node's name, eg `"cube"` will receive a collider that's named `"cube.HOTHAM_COLLIDER_WALL"`

- You can now define colliders with (mostly) arbitrary geometry 🎉 

## Future work
- https://github.com/leetvr/hotham/issues/363

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,14 @@ members = ["hotham", "hotham-simulator", "examples/simple-scene", "examples/crab
 [profile.dev.package.rapier3d]
 opt-level = 3
 
-[profile.dev.package.base64]
+[profile.dev.package.parry3d]
+opt-level = 3
+
+# Make performance critical packages compile with optimizations
+[profile.dev.package.nalgebra]
+opt-level = 3
+
+[profile.dev.package."base64"]
 opt-level = 3
 
 [profile.dev.package.gltf]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,11 +8,10 @@ opt-level = 3
 [profile.dev.package.parry3d]
 opt-level = 3
 
-# Make performance critical packages compile with optimizations
 [profile.dev.package.nalgebra]
 opt-level = 3
 
-[profile.dev.package."base64"]
+[profile.dev.package.base64]
 opt-level = 3
 
 [profile.dev.package.gltf]
--- a/hotham/src/asset_importer/mod.rs
+++ b/hotham/src/asset_importer/mod.rs
@@ -362,7 +362,7 @@ fn find_wall_collider_for_node<'a>(
 
     // Iterate through each node to try and find the matching node, then fetch its mesh.
     import_context.document.nodes().find_map(|n| {
-        if n.name()? != &wall_pattern {
+        if n.name()? == wall_pattern {
             return None;
         }
 
--- a/hotham/src/asset_importer/mod.rs
+++ b/hotham/src/asset_importer/mod.rs
@@ -362,7 +362,7 @@ fn find_wall_collider_for_node<'a>(
 
     // Iterate through each node to try and find the matching node, then fetch its mesh.
     import_context.document.nodes().find_map(|n| {
-        if n.name()? == wall_pattern {
+        if n.name()? != wall_pattern {
             return None;
         }
 
--- a/hotham/src/asset_importer/mod.rs
+++ b/hotham/src/asset_importer/mod.rs
@@ -3,19 +3,25 @@ pub mod scene;
 
 use crate::{
     components::{
-        animation_controller::AnimationController, GlobalTransform, Info, LocalTransform, Mesh,
-        Parent, Root, Skin, Visible,
+        animation_controller::AnimationController, Collider, GlobalTransform, Info, LocalTransform,
+        Mesh, Parent, Root, Skin, Visible,
     },
     rendering::{light::Light, material::Material},
-    resources::{RenderContext, VulkanContext},
+    resources::{physics_context::PhysicsContext, RenderContext, VulkanContext},
 };
 use anyhow::Result;
 
 use gltf::Document;
 use hecs::{Entity, World};
-use std::{borrow::Cow, collections::HashMap};
+use itertools::Itertools;
+use nalgebra::Isometry3;
+use std::{borrow::Cow, collections::HashMap, convert::TryInto};
 
 use self::scene::Scene;
+
+static COLLIDER_TAG: &str = ".HOTHAM_COLLIDER";
+static WALL_COLLIDER_TAG: &str = ".HOTHAM_COLLIDER_WALL";
+static SENSOR_COLLIDER_TAG: &str = ".HOTHAM_COLLIDER_SENSOR";
 
 /// Convenience type for models
 pub type Models = HashMap<String, World>;
@@ -24,6 +30,7 @@ pub type Models = HashMap<String, World>;
 pub(crate) struct ImportContext<'a> {
     pub vulkan_context: &'a VulkanContext,
     pub render_context: &'a mut RenderContext,
+    pub physics_context: &'a mut PhysicsContext,
     pub models: Models,
     pub node_entity_map: HashMap<usize, Entity>,
     pub mesh_map: HashMap<usize, Mesh>,
@@ -36,6 +43,7 @@ impl<'a> ImportContext<'a> {
     fn new(
         vulkan_context: &'a VulkanContext,
         render_context: &'a mut RenderContext,
+        physics_context: &'a mut PhysicsContext,
         glb_buffer: &'a [u8],
     ) -> Self {
         let glb = gltf::Glb::from_slice(glb_buffer).unwrap();
@@ -47,6 +55,7 @@ impl<'a> ImportContext<'a> {
         Self {
             vulkan_context,
             render_context,
+            physics_context,
             models: Default::default(),
             node_entity_map: Default::default(),
             mesh_map: Default::default(),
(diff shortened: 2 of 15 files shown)


hotham/src/asset_importer/mod.rs
```diff
@@ -176,6 +207,21 @@ fn load_models_from_gltf_data(import_context: &mut ImportContext) -> Result<()>
...
+        if !node.name().unwrap_or_default().contains(COLLIDER_TAG) {
+            continue;
+        }
+
+        if let Some(mesh) = node.mesh() {
+            mesh_ids.push(mesh.index())
+        }
+    }
+
+    mesh_ids
```
rasmusgo, line 222: Did you consider using `filter_map` for this?

hotham/src/asset_importer/mod.rs
```diff
@@ -229,18 +281,131 @@ fn load_node(
...
+
+    Some(Collider::new(handle))
+}
+
+fn find_wall_collider_for_node<'a>(
+    name: &'a str,
+    import_context: &'a ImportContext,
+) -> Option<(&'a str, gltf::Mesh<'a>)> {
+    // Create a pattern to search for the collider's name:
+    let wall_pattern = format!("{}.HOTHAM_COLLIDER_WALL", name);
```
rasmusgo, line 360: This duplication is error prone as it looks like the tag contents is defined only once but isn't actually.

hotham/src/asset_importer/mod.rs
```diff
@@ -229,18 +281,131 @@ fn load_node(
...
+
+    // Iterate through each node to try and find the matching node, then fetch its mesh.
+    import_context.document.nodes().find_map(|n| {
+        if !n.name()?.contains(&wall_pattern) {
+            return None;
+        }
+
+        n.mesh().map(|mesh| (n.name().unwrap(), mesh))
```
rasmusgo, line 368:
```suggestion
        if n.name()?.contains(&wall_pattern) {
            n.mesh().map(|mesh| (n.name().unwrap(), mesh))
        } else {
            None
        }
```

hotham/src/asset_importer/mod.rs
```diff
@@ -229,18 +281,131 @@ fn load_node(
...
+fn find_wall_collider_for_node<'a>(
+    name: &'a str,
+    import_context: &'a ImportContext,
+) -> Option<(&'a str, gltf::Mesh<'a>)> {
+    // Create a pattern to search for the collider's name:
+    let wall_pattern = format!("{}.HOTHAM_COLLIDER_WALL", name);
+
+    // Iterate through each node to try and find the matching node, then fetch its mesh.
+    import_context.document.nodes().find_map(|n| {
+        if !n.name()?.contains(&wall_pattern) {
```
rasmusgo, line 364: I think this will match more than it should. Consider `blue_cube.HOTHAM_COLLIDER_WALL` vs `cube.HOTHAM_COLLIDER_WALL`.

hotham/src/asset_importer/mod.rs
```diff
@@ -134,7 +155,17 @@ fn load_models_from_gltf_data(import_context: &mut ImportContext) -> Result<()>
...
     let mut animation_controller_entity = None;
-    for node in document.scenes().next().unwrap().nodes() {
+
+    // NOTE: We don't currently support multiple scenes, so we just take the first one.
+    let scene = document.scenes().next().unwrap();
+
+    // Iterate through each of the root nodes in the scene and load it in.
+    for node in scene.nodes() {
+        // Don't add wall collider geometry as nodes.
+        if node.name().unwrap_or_default().contains(WALL_COLLIDER_TAG) {
```
rasmusgo, line 165: You might want to use `ends_with` instead of `contains` here and elsewhere.

rasmusgo (changes requested):

hotham/src/asset_importer/mod.rs
```diff
@@ -123,7 +136,15 @@ fn load_models_from_gltf_data(import_context: &mut ImportContext) -> Result<()>
     // A bit lazy, but whatever.
     let document = import_context.document.clone();
 
+    // Identify meshes that will be used for collider geometry.
+    let collider_mesh_ids = get_collider_mesh_ids(document.nodes());
+
     for mesh in document.meshes() {
+        // Don't load meshes that are going to be used as collider geometry
+        if collider_mesh_ids.contains(&mesh.index()) {
```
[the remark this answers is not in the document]
kanerogers: Happy to optimise this in future! 😄

hotham/src/asset_importer/mod.rs
kanerogers: I did, but it started to look quite messy. 😦

hotham/src/asset_importer/mod.rs
kanerogers: I much prefer early returns to branches - way more readable

hotham/src/asset_importer/mod.rs
kanerogers: Ah damn, I thought I'd caught all those intances!

hotham/src/asset_importer/mod.rs
kanerogers: Ah, good pickup. It should just be a straight equals, actually!

hotham/src/asset_importer/mod.rs
rasmusgo: You could still swap the order of them to get rid of the double negation.

```suggestion
        if n.name()?.contains(&wall_pattern) {
            return n.mesh().map(|mesh| (n.name().unwrap(), mesh));
        }
        None
```

hotham/src/asset_importer/mod.rs
kanerogers: That's a good point! There's only one case where this doesn't work (when we're skipping importing a mesh that could be either type of collider), but it works everywhere else!

hotham/src/asset_importer/mod.rs
kanerogers: I think it's just not as clear that way - I like writing the "happy path" as the last statement of a branch

hotham/src/asset_importer/mod.rs
```diff
@@ -229,18 +281,132 @@ fn load_node(
...
+fn find_wall_collider_for_node<'a>(
+    name: &'a str,
+    import_context: &'a ImportContext,
+) -> Option<(&'a str, gltf::Mesh<'a>)> {
+    // Create a pattern to search for the collider's name, suffixed with the wall collider tag.
+    let wall_pattern = format!("{}{}", name, WALL_COLLIDER_TAG);
+
+    // Iterate through each node to try and find the matching node, then fetch its mesh.
+    import_context.document.nodes().find_map(|n| {
+        if n.name()? == wall_pattern {
```
rasmusgo, line 365: I think this got flipped the wrong way

hotham/src/asset_importer/mod.rs
```diff
@@ -176,6 +207,21 @@ fn load_models_from_gltf_data(import_context: &mut ImportContext) -> Result<()>
     Ok(())
 }
 
+fn get_collider_mesh_ids(nodes: gltf::iter::Nodes) -> Vec<usize> {
+    let mut mesh_ids = Vec::new();
+    for node in nodes {
+        if !node.name().unwrap_or_default().contains(COLLIDER_TAG) {
```
rasmusgo, line 213: `ends_with`

Cargo.toml
```diff
@@ -5,7 +5,14 @@ members = ["hotham", "hotham-simulator", "examples/simple-scene", "examples/crab
...
 
-[profile.dev.package.base64]
+[profile.dev.package.parry3d]
+opt-level = 3
+
+# Make performance critical packages compile with optimizations
+[profile.dev.package.nalgebra]
+opt-level = 3
+
+[profile.dev.package."base64"]
```
rasmusgo, line 15: Why are there quotes around base64 now and not before?

hotham/src/asset_importer/mod.rs
kanerogers: Oh yikes, it did too! Annoying that the test didn't pick that up..

hotham/src/asset_importer/mod.rs
kanerogers: Can't do `ends_with` here! We want to see if it's *either* a wall or sensor collider, `ends_with` wouldn't meet that test.

Cargo.toml
kanerogers: Ah, somebody had suggested this as a fix for getting cargo to optimise at the right level. Didn't do anything, so I'll revert that!

hotham/src/asset_importer/mod.rs
rasmusgo: Wouldn't it be better to test that instead then? This will let other things through as well.

rasmusgo: If there is a problem with the name, eg blender added a `.001` suffix, the mesh will be silently ignored here but never added as a collider. It would be good to add a test for that actually.

kanerogers: I think that's totally fine behaviour for now, honestly. This is just a first pass implementation that we can come back to in #351. 😄

rasmusgo (approved):
As a work in progress, this LGTM.

PR #371, opened 2022-09-27
Refactor `physics_system`

### LINKED ISSUE #366: [Maintenance] Refactor `physics_system`
## Background
At the moment, our integration with Rapier is a little loose. Developers need to do a lot of work to get their game simulations synchronized with the physics simulation. Let's try and fix that.

## Design
Almost every problem in game engine design is some sort of synchronization problem. The easiest way to simplify synchronization problems is to define the source of truth, ie. "who is authoritative over this data".

What data do we want to synchronize between the physics simulation and the game simulation? Mostly, it's just global transforms (ie. where is this entity located in global space).

At a high level, there's 3 ways an entity might want to have its global transform synchronized with the physics simulation:

1. No synchronization at all
1. Physics authoritative (ie. after the simulations have synchronized, `GlobalTransform` will be equal to the position in the physics simulation)
    - eg. objects whose movement we want modelled in some sort of physical way (ie. by simulating gravity or velocity etc).
1. Game authoritative (ie. after the simulations have synchronized, the position in the physics simulation will be equal to `GlobalTransform`).
    - eg. things like colliders which don't have any dynamics but still need a position in the physics simulation, or kinematic objects in the physics simulation

A simple `PhysicsControlled` tag component could be able to indicate to the relevant system what should be considered authoritative. It is still up to the game developer to add this tag to the right kinds of entities, and ensure their physics simulation representations are configured correctly.

## TODO
- [x] Create `PhysicsControlled` component
- [x] Handle case 1
- [x] Handle case 2

kanerogers:
## What is this related to?
Fixes #366

## What changed?
- Significantly simplified physics integration
- Cleaned up test suite

See #366 for more details


hotham/src/systems/audio.rs
```diff
@@ -121,7 +121,7 @@ mod tests {
     };
 
     use rapier3d::prelude::{RigidBodyBuilder, RigidBodyType};
-    const DURATION_SECS: u32 = 8;
+    const DURATION_SECS: u32 = 3;
```
rasmusgo, line 124: This looks unrelated

kanerogers: Just trimming the tests down! 😄 A bit unrelated, but since the whole test suite was running like kludge, I felt like this was fair.

kanerogers:
Thanks for all the discussion on this one @ickk and @rasmusgo! I'm going to go ahead and merge this one in now as I want to keep up the pace. Definitely want your feedback on how this feels outside the engine!

hotham/src/contexts/physics_context.rs
```diff
@@ -32,7 +33,12 @@ impl Default for PhysicsContext {
         let (contact_force_send, contact_force_recv) = crossbeam::channel::unbounded();
         let event_handler = ChannelEventCollector::new(collision_send, contact_force_send);
         let gravity: Matrix3x1<f32> = vector![0.0, 0.0, 0.0]; // TODO: no gravity in SPACE baby! But some games may uh, need this.
-        let integration_parameters = IntegrationParameters::default();
+        let integration_parameters = IntegrationParameters {
+            dt: 1. / 72.,
```
rasmusgo, line 37: This constant should probably be extracted from here.

kanerogers: Good point! :D I have done that now in #372

PR #373, opened 2022-09-28
Switch to using glam for maths

### LINKED ISSUE #372: [Maintenance] Replace `nalgebra` with `glam`
## Background
**[Nalgebra, I love you, but you're bringing me down](https://www.youtube.com/watch?v=-eohHwsplvY)**.

`nalgebra` is a great maths library. It does *lots* of cool stuff. 

_But,_ it's a pain in the fucking arse in most cases. So many things that are otherwise simple in other Maths libraries are just mind-bendingly complicated in `nalgebra`. Its errors are indecipherable, its documentation is a maze of darkness and it uses so many traits that it [gets specifically called out in the `cargo` docs for its heavy use of generics](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides-and-generics).

So, let's [hurl it into the sun](https://www.wired.com/2017/03/lets-physics-knocking-asteroid-sun/) and replace it with `glam`.

## TODO
- [x] How do we keep `rapier` and still have `nalgebra`?
- [ ] Do it

kanerogers:
## What is this related to?
Closes #372

## What changed?
Switched to `glam` as our primary maths library

Hotham is now simpler to use, maths related code is simpler and easier to reason about.


hotham/src/rendering/camera.rs
```diff
@@ -1,46 +1,112 @@
...
         self.view_from_gos
     }
 
     /// Get the camera's position
-    pub fn position_in_gos(&self) -> Vector4<f32> {
-        let p = self.gos_from_view.translation.vector;
-        vector![p[0], p[1], p[2], 0.]
+    pub fn position_in_gos(&self) -> Vec4 {
+        let p = self.gos_from_view.translation;
+        [p[0], p[1], p[2], 0.].into()
```
rasmusgo, line 38: This is an old thing and not very important since it is never read but this vector should end with a 1. It will not transform correctly when multiplied with matrices otherwise.

kanerogers: Ah, good pickup!

hotham/src/systems/pointers.rs
```diff
@@ -54,7 +36,10 @@ pub fn pointers_system_inner(
     let global_from_stage = stage::get_global_from_stage(world);
 
     // Create a transform from local space to grip space.
-    let grip_from_local = ROTATION_OFFSET * POSITION_OFFSET;
+    // NOTE: This is most likely *WRONG* as the order for these transforms was not recoreded correctly.
+    // TODO: Make these correct.
+
+    let grip_from_local = Affine3A::from_rotation_translation(ROTATION_OFFSET, POSITION_OFFSET);
```
rasmusgo, line 42: I think you should be able to do something like
```suggestion
    let grip_from_local = Affine3A::from_rotation_translation(ROTATION_OFFSET, Vec3::ZERO) * 
        Affine3A::from_rotation_translation(Quat::IDENTITY, POSITION_OFFSET);
```

kanerogers: Agh, rats, I forgot to update the comment! I ended up just asking `nalgebra` what the correct values _should_ be and updated them accordingly

rasmusgo: It may still make more sense to express them in a more natural order where the offset is along a single axis.

rasmusgo: It may even make more sense to express the offset as `local_from_grip` and then invert it.

kanerogers: Good thinking! Happy to come back to this later though, it's not the end of the world

PR #375, opened 2022-09-29
[Maintenance] Fix simulator camera

kanerogers:
## Background
- Fixes rotations on the simulator camera
- Decreases the field of view to be closer to what's seen inside the headset (the helmet is actually really, really big)


kanerogers:
This is a smol one; I'll merge it in.

Issue #368, opened 2022-09-26
[Feature] Simplify artificial locomotion


kanerogers:
## Background
Artificial locomotion is pretty tricky at the moment. Let's come up with a simpler way of achieving that.

kanerogers:
Closed by #376

PR #376, opened 2022-09-30
Simplify virtual locomotion

kanerogers:
## What is this related to?
#368

## What changed?
- Adds a `HMD` component to keep track of the user's headset
- The engine is responsible for adding both `Stage` and `HMD` entities on startup
- The `HMD` is parented to the `Stage`
- The engine updates the `LocalPosition` of the `HMD` each frame
- This makes life super simple - the `HMD`'s `GlobalTransform` is its position in global space, its `LocalTransform` is its position in stage space

Artifical locomotion is now much easier!


hotham/src/contexts/input_context.rs
```diff
@@ -322,6 +322,28 @@ impl RightInputContext {
     }
 }
 
+#[derive(Debug, Default)]
+/// Input from the Head Mounted Display (HMD, or headset)
+pub struct HmdInputContext {
+    left_eye_in_stage: Affine3A,
+    right_eye_in_stage: Affine3A,
```
rasmusgo, line 329: This is a transform, not a point, so it should be `right_eye_from_stage` or `stage_from_right_eye`.

hotham/src/asset_importer/mod.rs
```diff
@@ -327,25 +328,39 @@ fn get_collider_for_node(
...
+        rapier3d::geometry::ColliderBuilder::new(shape).user_data(this_entity.to_bits().get() as _);
 
     // If this is a wall collider, ensure it's not a sensor.
     let collider = if collider_node_name.ends_with(WALL_COLLIDER_TAG) {
         println!(
             "[HOTHAM_ASSET_IMPORTER] Created wall collider for model {}",
             collider_node_name
         );
-        collider_builder.sensor(false).build()
+        collider_builder
```
kanerogers, line 345: These fixes to the collider import were necessary to test locomotion correctly.

hotham/src/systems/rendering.rs
```diff
@@ -289,14 +289,17 @@ mod tests {
         let (_, mut world) = models.drain().next().unwrap();
 
         // Add stage transform
-        let stage_entity = stage::add_stage(&mut world, &mut physics_context);
-        let mut stage_local_transform = world.get_mut::<LocalTransform>(stage_entity).unwrap();
-        stage_local_transform.translation = [0.1, 0.2, 0.3].into();
-        stage_local_transform.rotation =
-            Quat::from_scaled_axis(Vec3::Y * (std::f32::consts::TAU * 0.1));
-
+        let stage_local_transform = LocalTransform {
```
kanerogers, line 292: While working more with the stage entity I was able to simplify this test a bit

hotham/src/contexts/input_context.rs
```diff
@@ -322,6 +322,28 @@ impl RightInputContext {
...
+        // Since engine will call `update_views()` *just before* calling this method, we
+        // can be sure that this data is up-to-date.
+        let views = &xr_context.views;
+        self.left_eye_in_stage = affine_from_posef(views[0].pose);
+        self.right_eye_in_stage = affine_from_posef(views[1].pose);
+    }
+
+    /// The position of the HMD in the real world (stage space)
+    pub(crate) fn position_in_stage(&self) -> Affine3A {
+        lerp_slerp(&self.left_eye_in_stage, &self.right_eye_in_stage, 0.5)
```
rasmusgo, line 343: OpenXR can give you this directly if you want.

hotham/src/contexts/input_context.rs
```diff
@@ -331,12 +353,13 @@ impl RightInputContext {
...
     pub left: LeftInputContext,
     pub right: RightInputContext,
+    pub hmd: HmdInputContext,
 }
 
 impl InputContext {
     /// Synchronize the context state with OpenXR. Automatically called by `Engine`
     /// each tick.
-    pub fn update(&mut self, xr_context: &XrContext) {
+    pub(crate) fn update(&mut self, xr_context: &XrContext) {
```
kanerogers, line 362: Developers should never have to call this method.

hotham/src/engine.rs
```diff
@@ -85,23 +86,44 @@ impl<'a> EngineBuilder<'a> {
             .expect("!!FATAL ERROR - Unable to initialize renderer!");
         let gui_context = GuiContext::new(&vulkan_context);
 
+        // Initialize the world with our "tracking" entities, the stage and the HMD.
+        let mut world = hecs::World::default();
+        let (stage_entity, hmd_entity) = create_tracking_entities(&mut world);
```
kanerogers, line 91: Since we treat `Stage` as a first class entity (ie. our renderer knows about it!) it makes sense to have the engine create this on startup and keep a reference to the entity to avoid having to keep query for the stage all the time.

rasmusgo (changes requested):

rasmusgo:
I think this is generally a good direction to go!

hotham/src/contexts/input_context.rs
kanerogers: I disagree - it is a point. Nothing is ever transformed from that space, and it makes it _very clear_ what this represents.

rasmusgo: Position does not include orientation. It is very confusing to overload well established terms, see https://en.wikipedia.org/wiki/Position_(geometry). "Pose" is a better term if you want to avoid "transform", it is also used heavily within OpenXR. Yes, I know rapier uses "position" instead of pose for some reason but that does not mean it is a wide spread use. Let's take a more thorough look:

**OpenXR** uses pose to mean orientation and position. https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XrPosef a pose has a XrVector3f position.

**Unity** separates position from rotation and scale and calls it pose or transform when combined:
https://docs.unity3d.com/ScriptReference/Rigidbody.html - a rigid body has a position and a rotation. You can also get this as a transform, https://docs.unity3d.com/ScriptReference/Transform-position.html - a transform has a Vector3 position. https://docs.unity3d.com/ScriptReference/Pose.html - a pose has a Vector3 position.

**Unreal** separates position and orientation and calls it pose or transform when combined:
https://docs.unrealengine.com/5.0/en-US/BlueprintAPI/SteamVR/GetTrackedDevicePositionandOrien-

**Bevy** uses Transform consistently in the code but sometimes writes position in the docs: https://docs.rs/bevy/latest/bevy/transform/components/struct.Transform.html

**Godot** uses transform consistently.https://docs.godotengine.org/en/stable/tutorials/math/vector_math.html#doc-vector-math and https://docs.godotengine.org/en/stable/tutorials/3d/using_transforms.html#transforms-are-your-friend

**Blender** uses transform with location, rotation, scale. It uses pose for skeletal rigging and animations.

But **rapier** uses position to mean both translation and orientation 🤦 https://rapier.rs/docs/user_guides/rust/rigid_bodies/#position

kanerogers: Okay, that is some *next level* research. I really, really appreciate that @rasmusgo. I am now convinced that position is not the best name. 😄 

`left_eye_in_stage` is a good compromise, that plus the type and its usage makes it very easy. We can add a doc string if we're feeling crazy. 

I'm currently on holiday, so if you could please make the suggested change and approve the PR, I'll merge it in. :)

hotham/src/components/stage.rs
```diff
@@ -1,39 +1,32 @@
...
+/// On startup, [`crate::Engine`] will create an entity with the [`Stage`] component, and an entity with the
+/// [`super::HMD`] component (representing the user's headset), parented to the stage entity.
+///
+/// This makes moving the player around in the virtual world *relatively* straightforward, as all is required is
+/// to move the [`Stage`]. The engine will ensure that the player's position in the game simulation AND the
+/// virtual cameras are correctly updated to account for this.
+///
+/// In short, the final position of the player in the game simulation (ie. global space) is:
+///
+/// `stage.position * hmd.position`
```
rasmusgo, line 12: I am a bit uncertain about what "position" should be replaced with here. "pose" or "transform"?

rasmusgo (approved):

rasmusgo:
I didn't realize that auto-merge was enabled. Oh well 🤷

kanerogers:
Oh well! Thanks for this. :-)

PR #389, opened 2022-10-13
Quadrics geometry in complex scene

rasmusgo:
I want to add quadrics as an example of a custom shader in the complex scene example. This code does just that. However, it modifies the Hotham engine in order to do so. I would like some help figuring out how to refactor this code to minimize changes to the engine.


kanerogers:
This is looking good!

I think that this is overloading Complex Scene a bit as it's not really an example of a complex scene anymore; it's an example of custom rendering. 

Could you rename this to something like `custom_rendering_example` leaving `complex_scene` as it was? I think more examples are a good thing, and complex scene is a good standalone example!

rasmusgo:
This has been superseded by #400.

PR #391, opened 2022-10-17
Make physics system entirely ECS driven

### LINKED ISSUE #379: [Feature] Make `PhysicsContext` ECS driven
## TODO
- [ ] Change PhysicsContext to put all physics related properties into the RigidBody and Collider components.
- [ ] Porting all of Rapier in there is going to be a little tedious, so we can just start with what's used in Hotham games today (eg. Crab Saber and The Station).

## Wrinkles
Queries like raycasting etc. don't quite make sense with components. We could come up with a nicer wrapper that combines world.query so that you could eg. cast a ray and then get back the components of an entity you hit.

kanerogers:
## What is this related to?
Closes #379

## What changed?
- Handles to Rapier are now private to the physics system
- Cleaned up examples to use new physics system
- Updated complex example to use two helmets

- Developers can use the `RigidBody` or `Collider` components to add physics functionality to their entities like so:

```rust
world.insert_one(entity, RigidBody { body_type: BodyType::Dynamic, velocity: [1., 0., 0.].into(), ..Default::default() })
```

- Sometimes it's necessary to tell the physics system to do something *just once*, for example, adding an impulse or teleporting a dynamic rigid-body. To help make this easier, you can add a "one shot" component like so. This component will be processed and then immediately removed

```rust
world.insert_one(entity, Teleport {})
```

## Follow ups
We don't have a way of cleaning up unused colliders/rigid-bodies yet. A simple approach might be to iterate through rapier's list of rigid-bodies and colliders and see if it can find the corresponding entity from its `user_data`


hotham/src/systems/rendering.rs
```diff
@@ -397,7 +389,10 @@ mod tests {
         views: &Vec<openxr::View>,
     ) -> Result<(), String> {
         // Render the scene
-        let mut renderdoc = begin_renderdoc();
+
```
kanerogers, line 392: I noticed that sometimes when running the test suite I would get a crash somewhere outside of Rust. I was thinking that it was something GPU related but after running the test suite in the debugger, it seems like there Renderdoc was crashing somewhere in its Vulkan Validation layers. It seems to only happen when running the entire test suite - not when running this test in isolation - so I've commented out the Renderdoc functionality for now.

hotham/src/components/local_transform.rs
```diff
@@ -76,8 +76,19 @@ impl LocalTransform {
         (self.scale, self.rotation, self.translation) = transform.to_scale_rotation_translation();
     }
 
+    /// Update ONLY rotation and rotation from a [`glam::Affine3A`]
+    pub fn update_rotation_translation_from_affine(&mut self, transform: &glam::Affine3A) {
```
kanerogers, line 80: This is another halfway solution to having a specific type that only contains rotation and translation.

kanerogers:
Merged with BDFL powers!

PR #392, opened 2022-10-17
Fix grabbing sensitivity

rasmusgo:
Use the same threshold for grabbing as for `grip_button` in `InputContext`. This makes the `grabbing_system` work via link.


kanerogers (approved):
Good pickup!

PR #397, opened 2022-10-25
[Rendering] Handle AO correctly

### LINKED ISSUE #395: [Rendering] Handle AO correctly
## Background
The [glTF spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material) states that:

> The textures for metalness and roughness properties are packed together in a single texture called metallicRoughnessTexture. Its green channel contains roughness values and its blue channel contains metalness values. This texture MUST be encoded with linear transfer function and MAY use more than 8 bits per channel.

Which we've handled to date by using ASTC's dual channel (LA) mode. However, the spec continues to talk about ambient occlusion textures later (emphasis mine)

> The occlusion texture; it indicates areas that receive less indirect lighting from ambient sources. Direct lighting is not affected. The **red** channel of the texture encodes the occlusion value, where 0.0 means fully-occluded area (no indirect lighting) and 1.0 means not occluded area (full indirect lighting). Other texture channels (if present) do not affect occlusion.

This very careful separation of red/blue/green between AO/metalness/roughness is deliberate. Most glTF exporters pack all three textures into a single image, such as the [Blender exporter](https://docs.blender.org/manual/en/2.80/addons/io_scene_gltf2.html#baked-ambient-occlusion). 

As such, we should treat the metalicRoughness texture as having **three** channels and not perform any swizzling.

kanerogers:
## What is this related to?
Fixes #395

## What changed?
We now import AO textures correctly

## What's the impact
Models with combined metallic roughness and AO textures (which is the common case) will see a significant performance improvement

### DIFF
--- a/hotham/src/asset_importer/mod.rs
+++ b/hotham/src/asset_importer/mod.rs
@@ -645,7 +645,7 @@ mod tests {
                     assert_eq!(
                         (
                             material.base_color_texture_set,
-                            material.physical_descriptor_texture_id,
+                            material.metallic_roughness_texture_id,
                             material.normal_texture_set,
                             material.occlusion_texture_set
                         ),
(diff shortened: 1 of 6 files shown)


rasmusgo (approved):
Good stuff!

PR #399, opened 2022-10-26
Fix incorrect f0 term in fragment shader

### LINKED ISSUE #393: [Rendering] Support loading tangents from assets again
## Background
We've gone back and forth on tangents a couple times; @rasmusgo found that computing tangents on the fly is actually quicker, so we've implemented this in https://github.com/leetvr/hotham/pull/342

*However*, some models simply require tangents that were calculated offline, eg.

image
image

kanerogers:
## What is this related to?
Fixes #393

## What changed?
Fixed a bug in the fragment shader, extracted magic number to avoid future issues.

PBR shader now looks much more correct:

### Before:
image

### After:
image

### DIFF
--- a/hotham/src/systems/debug.rs
+++ b/hotham/src/systems/debug.rs
@@ -27,7 +27,7 @@ pub fn debug_system(engine: &mut Engine) {
 
     if input_context.right.a_button_just_pressed() {
         let params = &mut render_context.scene_data.params;
-        params.x = ((params.x - 0.1) % 5.) as f32;
+        params.x = ((params.x + 5. - 0.1) % 5.) as f32;
         println!("[HOTHAM_DEBUG] params.x is now {}", params.x);
     }
 }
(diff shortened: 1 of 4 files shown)


hotham/src/systems/debug.rs
```diff
@@ -24,4 +24,10 @@ pub fn debug_system(engine: &mut Engine) {
         params.x = ((params.x + 0.1) % 5.) as f32;
         println!("[HOTHAM_DEBUG] params.x is now {}", params.x);
     }
+
+    if input_context.right.a_button_just_pressed() {
+        let params = &mut render_context.scene_data.params;
+        params.x = ((params.x - 0.1) % 5.) as f32;
```
rasmusgo, line 30: This can turn negative. I think you want to something like this: 
```suggestion
        params.x = ((params.x + 5. - 0.1) % 5.) as f32;
```

rasmusgo (approved):
Looks good from what I can tell without running it.

PR #401, opened 2022-10-30
Fix rendering test

rasmusgo:
Fix output path of rendered images and update them to reflect the recent changes made in #399.


kanerogers (approved):
Thanks for this! Good pickup

PR #402, opened 2022-10-31
Use float16 in fragment shader

rasmusgo:
~Almost working as it should. There is a problem with specular highlight that I am trying to pin down at the moment.~

~I also made a convenience thing for updating the pbr rendering tests if `UPDATE_IMAGES` env var is truthy. That might belong in a separate PR.~

I have fixed several issues with the spotlights and I think this is starting to look good enough now. There is a difference with rendering reflections of spotlights on very shiny surfaces. The new code is more likely to show a reflection of the spotlight but it it flickers. This is visible in the normal tangents test scene but not on the helmet.

I ran the many helmets benchmark but was surprised to see no difference in performance. Perhaps the bottleneck is somewhere else?

PR #403, opened 2022-10-31
Add render normal tangent test

rasmusgo:
This new test catches the problem with specular highlight in #402.

### DIFF
--- a/hotham/Cargo.toml
+++ b/hotham/Cargo.toml
@@ -41,7 +41,6 @@ renderdoc = "0.10"
 
 [dev-dependencies]
 approx = "0.5"
-lenient_bool = "0.1.1"
 serde_json = "1.0"
 
 [target.'cfg(target_os = "android")'.dependencies]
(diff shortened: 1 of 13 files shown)


hotham/Cargo.toml
```diff
@@ -41,6 +41,7 @@ renderdoc = "0.10"
 
 [dev-dependencies]
 approx = "0.5"
+lenient_bool = "0.1.1"
```
kanerogers, line 44: New dependency alert! Isn't is sufficient to just check to see if the environment variable is defined? Who cares what the value is.

rasmusgo: Users of hotham will not be impacted because it's only a dev decency. It felt more proper to respect the value of the environment variable but it doesn't do much, if any, difference to the current usage. I can remove it if you feel strongly about it.

rasmusgo: You don't even need to feel strongly about it. It was easy to remove and replace it with a test for the truthy strings.

kanerogers: Thank you! Yeah because it's just a testing thing I felt like actually checking the value of the var was overkill, but this is a good solution 👌

kanerogers (approved):
Nice work. Really like having better tests for the renderer.

PR #405, opened 2022-11-06
Upgrade openxr to 0.17

rasmusgo:
Hotham fails to build with rust 1.65 with an error in openxr. Updading openxr solves this.


rasmusgo:
Tested natively on the quest and with the simulator on the simple scene example.

kanerogers (approved):
LGTM on 1.65 on simulator and headset! Thanks for taking care of this.

Issue #364, opened 2022-09-26
[Maintenance] Standardise naming of binaries in examples


kanerogers:
## Background
At the moment, each of our examples is called.. lots of different things. Sometimes that's to get around conflicts with targets (eg. the lib is built as `simple_scene` so the binary can't be called `simple_scene`), other times it's just my laziness.

## Solution
Simple naming convention:

- The `lib` target is just called by the project name (eg. `simple_scene`).
- The `bin` target is the project name, with a `_desktop` suffix (eg. `simple_scene_desktop`)

To change the name of a `bin` target, just modify the `Cargo.toml` manifest to add / change the following:

```toml
[[bin]]
name = "simple_scene_desktop"
path = "src/main.rs"
```

PR #406, opened 2022-11-09
standartize example package's target names

Nikolay186:


rasmusgo:
This relates to #364. The names in the issue description does not have the `hotham_` prefix. @kanerogers do you want to remove or keep the prefix?

examples/simple-scene/Cargo.toml
```diff
@@ -6,9 +6,10 @@ version = "0.2.0"
 
 [lib]
 crate-type = ["lib", "cdylib"]
+name = "hotham_simple_scene_example"
 
 [[bin]]
-name = "hotham_simple_scene_example"
+name = "hotham_simple_scene_example_desktop"
```
rasmusgo, line 12: The suggested name in the issue description #364 is `simple_scene`.

Nikolay186: fixed in [this commit](https://github.com/leetvr/hotham/pull/406/commits/2498d02b2cf33e8100afaac2901e18b5de2c932b)

kanerogers:
Thanks so much for taking care of this, Nikolay! I’m just a bit swamped
this week, but will come back to it properly shortly. :-)

examples/simple-scene/Cargo.toml
kanerogers: I think the `hotham_` prefix is the way to go! Sorry, I'll clarify that in the issue. 😄

louiidev: > I think the `hotham_` prefix is the way to go! Sorry, I'll clarify that in the issue. 😄

curious, what's the thinking behind the prefix? feels more ergonomic to me to have it as `simple_scene_example_desktop`

PR #407, opened 2022-11-13
Simplify quadrics in custom rendering example

rasmusgo:
The code gets easier to read and possibly faster to run by doing less in the vertex shader and consolidate the logic in the fragment shader. It was also possible to simplify the fragment shader. The amount of data passed from the vertex shader to the fragment shader is greatly reduced. This makes vertices cheaper which is a bit funny when the point of this rendering technique is to enable perfectly curved geometry with very few vertices.

### DIFF
--- a/examples/custom-rendering/scripts/run_on_device.sh
+++ b/examples/custom-rendering/scripts/run_on_device.sh
@@ -8,4 +8,10 @@ cd $scriptdir/..
 
 cargo apk run --release
 
+# Wait for the app to start
+for i in 1 2 3 4 5; do
+    adb shell pidof rust.crab_saber && break
+    sleep 1
+done
+
 adb logcat --pid="$(adb shell pidof rust.custom_rendering_example)"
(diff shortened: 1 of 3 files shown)


examples/custom-rendering/scripts/run_on_device.sh
```diff
@@ -8,4 +8,10 @@ cd $scriptdir/..
 
 cargo apk run --release
 
+# Wait for the app to start
+for i in 1 2 3 4 5; do
+    adb shell pidof rust.crab_saber && break
```
kanerogers, line 13: Looking for crabsaber? 😛

kanerogers (approved):
That is absolutely fascinating. Those registers must be *very* expensive.

examples/custom-rendering/scripts/run_on_device.sh
rasmusgo: 🤦

PR #414, opened 2022-12-27
Push constants for materials

rasmusgo:
This is a subset of the `krperf` branch with several fixes added to fix stuff that was broken. There is at least one problem remaining: The desktop version of the helmet model has ambient occlusion stored in a separate texture but the support for that is dropped. This breaks the pbr rendering test.

This follows #413 and replaces #411. 

Before merging:
- [x] Update helmet model to fix ambient occlusion on desktop.
- [x] Ensure render tests pass, updating images as needed.


kanerogers:
I've updated the textures, but I'm getting some validation warnings:

```
VUID-vkCmdDrawIndexed-None-02699(ERROR / SPEC): msgNum: -1539028524 - Validation Error: [ VUID-vkCmdDrawIndexed-None-02699 ] Object 0: handle = 0x980b0000000002e, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0xa44449d4 | Descriptor set VkDescriptorSet 0x980b0000000002e[] encountered the following validation error at vkCmdDrawIndexed time: Descriptor in binding #3 index 5 is being used in draw but has never been updated via vkUpdateDescriptorSets() or a similar call. The Vulkan spec states: Descriptors in each bound descriptor set, specified via vkCmdBindDescriptorSets, must be valid if they are statically used by the VkPipeline bound to the pipeline bind point used by this command (https://vulkan.lunarg.com/doc/view/1.3.211.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDrawIndexed-None-02699)
    Objects: 1
        [0] 0x980b0000000002e, type: 23, name: NULL
```

kanerogers:
I've fixed all the validation warnings except this one:

```
UNASSIGNED-CoreValidation-Shader-OutputNotConsumed(WARN / PERF): msgNum: 101294395 - Validation Performance Warning: [ UNASSIGNED-CoreValidation-Shader-OutputNotConsumed ] Object 0: handle = 0xea7170000000031, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x609a13b | vertex shader writes to output location 0.3 which is not consumed by fragment shader
    Objects: 1
        [0] 0xea7170000000031, type: 15, name: NULL
```

As far as I can tell, this is wrong. Either:

- `0.3` refers to `layout location = 3` in which case it's *totally* wrong, because no vertex shader writes to `location = 3`
- `0.3` refers to `layout location = 2`, in which case it's also wrong because the fragment shader reads from `location = 2`

kanerogers (approved):
Looking good! Thanks so much for putting this together!

Issue #356, opened 2022-09-20
[Rendering] Use 16bit data types on the GPU where possible


kanerogers:
## Background
As Hotham's renderer is primarily memory bandwidth bound on the Quest 2, dropping to 16 bit data types will not only improve bandwidth, but in general only use a single instruction, compared to two, when operating on them. This could be a very significant performance increase.

## TODO
- [ ] Modify `Vertex` struct
- [ ] Modify asset importer to convert to the correct precision
- [ ] Modify shaders

kanerogers:
Closed in #415

Issue #274, opened 2022-07-19
[Feature] Implement fixed-foveated rendering


kanerogers:
## Background
Speaking with an Oculus Engineer, one of the *biggest* ways we can get an improvement out of our rendering pipeline is to use Fixed-Foveated rendering.

Facebook has an [OpenXR extension](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_foveation_vulkan) that should make this very easy to implement.

kanerogers:
Implementing FFR seems to need a couple of dependent extensions in OpenXR

- [`XR_FB_swapchain_update_state`](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_swapchain_update_state)
- [`XR_FB_foveation`](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_foveation)
- [`XR_FB_foveation_configuration`](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_foveation_configuration)
- [`XR_FB_foveation_vulkan`](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_foveation_vulkan)

And then on the Vulkan side:

- [`VK_EXT_fragment_density_map2`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_fragment_density_map2.html)

The setup seems to look something like this:

1. Create a `xrFoveationProfileFB` with `xrCreateFoveationProfileFB`
1. Call `xrUpdateSwapchainFB` with the `XrSwapchainStateFoveationFB` structure.

kanerogers:
This was closed in #415

PR #415, opened 2023-01-19
Rendering Optimisations - Part 2

### LINKED ISSUE #259: [Rendering] Rendering optimisations v2
## Background
After getting significant gains in #226, there's still much more juice we can squeeze out of the renderer.

The name of the game is _bandwidth_. We **must** ensure that we pull as little data down as possible in each draw.

## TODO
- [ ] #357
- [x] Use instancing (Qualcomm recommends this: https://developer.qualcomm.com/sites/default/files/docs/adreno-gpu/developer-guide/gpu/best_practices_vertex.html#consider-geometry-instancing)
- [ ] #356
- [ ] #274 
- [ ] https://www.unrealengine.com/en-US/blog/physically-based-shading-on-mobile
- [ ] https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_transform/README.md

kanerogers:
## Background
Closes #259, #357, #356, #274

This is a *very* big PR. It touches almost all of Hotham's internals, and makes many breaking changes.

However, it also *dramatically* improves the performance of almost all Hotham applications. While I would have preferred to split this PR into multiple chunks, strong time pressure and the nature of inter-connected improvements made that incredibly difficult.

## TODO
- [x] Run through the full test suite and create issues for any problems that arise
- [x] Test all examples on-device


kanerogers:
First bug is with FFR - there's no support for this in the simulator, obviously. I could potentially add some support for this, but it really doesn't make much sense. Instead I'll see if I can make it optional behind a `target_os` flag

kanerogers:
I've *mostly* got this working:

- All tests pass
- Crab Saber, Simple Scene and friends all work fine on simulator

Next up is testing on-device and looking for performance regressions!

kanerogers:
Android now builds! Just needs to be verified.

kanerogers:
Okay! Everything works fine, no performance regressions on The Station. All is well! 🎉

rasmusgo:
The gui render doesn't look good. The color of the ferris crab is broken, see `test_assets/render_draw_gui.jpg`
image

Cargo.toml
```diff
@@ -1,13 +1,15 @@
 [workspace]
 members = [
-    "hotham",
-    "hotham-simulator",
-    "examples/shared",
-    "examples/simple-scene",
-    "examples/complex-scene",
-    "examples/custom-rendering",
```
rasmusgo, line 8: Custom rendering example was removed by mistake here.

rasmusgo: And doesn't compile if added...

hotham/src/contexts/physics_context.rs
```diff
@@ -20,9 +19,9 @@ pub struct PhysicsContext {
     pub narrow_phase: NarrowPhase,
     pub rigid_bodies: RigidBodySet,
     pub island_manager: IslandManager,
-    pub collision_recv: Receiver<CollisionEvent>,
-    pub contact_force_recv: Receiver<ContactForceEvent>,
-    pub event_handler: ChannelEventCollector,
+    // pub collision_recv: Receiver<CollisionEvent>,
+    // pub contact_force_recv: Receiver<ContactForceEvent>,
+    // pub event_handler: ChannelEventCollector,
```
rasmusgo, line 24: Why is this removed?

rasmusgo:
There is dark arc on the flat area close to the spheres on the side towards the light in the normal tangents test. It looks weird and wasn't there before. Can it have anything to do with ambient occlusion? Spotlights should not be affected by ambient occlusion...

exporters/export.py
```diff
@@ -0,0 +1,40 @@
...
+# I will face Python.
+# I will permit it to pass over me and through me.
+# And when it has gone past, I will turn the inner eye to see its path.
+# Where the Python has gone there will be nothing.
+# Only Rust will remain
+
+class SaveOnWrite:
+    def export(dummy):
+        print("Hello?")
+        bpy.ops.export_scene.gltf(export_format="GLB", filepath="C:\\Users\\kanem\\Development\\hotham\\test_assets\\hot_cube.glb")
```
rasmusgo, line 27: This hardcoded path is not great. Both the path to the repo and the filename are problematic.

hotham-asset-client/src/client.rs
```diff
@@ -0,0 +1,218 @@
...
+
+pub async fn watch(asset_names: Vec<String>, sender: Sender<AssetUpdatedMessage>) -> Result<()> {
+    run_client(asset_names, sender).await
+}
+
+async fn run_client(
+    mut asset_names: Vec<String>,
+    sender: Sender<AssetUpdatedMessage>,
+) -> Result<()> {
+    let server_addr = "192.168.1.124:5000".parse().unwrap();
```
rasmusgo, line 19: This hardcoded address is problematic. Can we use a parameter or an environment variable instead?

hotham/src/contexts/physics_context.rs
```diff
@@ -1,4 +1,3 @@
-use crossbeam::channel::Receiver;
```
rasmusgo, line 1: This can be replaced with `rapier3d::crossbeam` if we don't want to depend on `crossbeam` directly.

rasmusgo:
It doesn't look like `hotham/src/shaders/all_textures_and_ibl.frag` or `hotham/src/shaders/all_textures_no_ibl.frag` are ever used, are they?

kanerogers:
> The gui render doesn't look good. The color of the ferris crab is broken, see `test_assets/render_draw_gui.jpg` image

Ah! Good pickup. I forgot to check these in on my desktop machine - have fixed these now.

Cargo.toml
kanerogers: Hrm, let me fix that up.

hotham/src/contexts/physics_context.rs
kanerogers: Good question!

- We never actually use Rapier events. We probably should, at some point in the future, because they're more efficient
- By removing them we can avoid an unnecessary crossbeam dependency

kanerogers:
> 



> It doesn't look like `hotham/src/shaders/all_textures_and_ibl.frag` or `hotham/src/shaders/all_textures_no_ibl.frag` are ever used, are they?

Nope! They're redundant - I'll kill 'em with fire.

exporters/export.py
kanerogers: Yeah, absolutely. This script was really just an experiment - I'll kill it off for now.

hotham-asset-client/src/client.rs
kanerogers: Yeah I went back and forth on a couple of approaches for this. We're missing a good general solution for configuration settings like these!

For now I've stuck to `option_env` which should check on the host machine for an environment variable. I made this optional as it would be extremely annoying to make this mandatory, and the error message will bubble back up to the engine if anybody is looking. I also made it a *host* variable as setting environment variables on Android is possible, but annoying (again, it would be good to have a good general solution for this).

kanerogers:
> There is dark arc on the flat area close to the spheres on the side towards the light in the normal tangents test. It looks weird and wasn't there before. Can it have anything to do with ambient occlusion? Spotlights should not be affected by ambient occlusion...
> 
> 

- re: ambient occlusion, I had to check the spec, but you are right! That's a bug - I've created an issue for it #416 
- But that's not an issue, as there's no ambient occlusion texture on that model
- Instead I think the issue is likely related to the (significantly faster) approximations we're using for the BRDFs, particularly the diffuse term, which I've mostly lifted them from Filament. For a real world scene, I haven't noticed any discernable difference, but given that we were originally using *the most accurate* BRDFs we could (lifted from a model viewer, rather than a real time engine), we will absolutely see some drop in accuracy in exchange for performance.
- I think this will always be a moving target between performance / quality. In the future it'd be good to give developers some more knobs they can tweak, but for now I believe this is an acceptable trade-off!

kanerogers:
Strangely, clippy in CI seems to be applying more lints than on my local machine. This is *very frustrating*..

kanerogers:
done

rasmusgo:
The custom rendering example is broken 😬

Cargo.toml
```diff
@@ -1,13 +1,16 @@
...
-    "examples/crab-saber",
-    "benchmarks/stress-test",
+  "hotham",
+  "hotham-simulator",
+  "examples/custom-rendering",
+  "examples/shared",
+  "examples/complex-scene",
+  "examples/simple-scene",
+  "examples/complex-scene",
```
rasmusgo, line 9: `complex-scene` is listed twice

rasmusgo (approved):

PR #417, opened 2023-01-28
Appease clippy for rust 1.67

rasmusgo:
These fixes are mostly about how local variables are formatted with `println!` and similar.


kanerogers (approved):
Thanks for running these on the tests too!

LGTM

PR #421, opened 2023-02-08
First pass at editor GUI

### LINKED ISSUE #419: Figure out how to draw a toolbar and inspector alongside the game

kanerogers:
## What is this related to?
Closes #419 

## What's changed?
- Began work on new, amazing, perfect and flawless OpenXR runtime (TM)
- Created stub app for editor
- Created a "editor" feature flag to keep @rasmusgo hap^H^H^H^H^H^H^Hencourage good development practices


hotham-editor/src/main.rs
```diff
@@ -0,0 +1,362 @@
...
+            let semaphore = device
+                .create_semaphore(
+                    &vk::SemaphoreCreateInfo::builder().push_next(&mut external_semaphore_info),
+                    None,
+                )
+                .unwrap();
+
+            let handle = external_semaphore
+                .get_semaphore_win32_handle(
+                    &vk::SemaphoreGetWin32HandleInfoKHR::builder()
```
rasmusgo, line 175: 😨 only windows?

hotham/src/contexts/xr_context/mod.rs
```diff
@@ -464,6 +464,10 @@ fn enable_xr_extensions(required_extensions: &mut xr::ExtensionSet) {
 #[cfg(not(target_os = "android"))]
 fn enable_xr_extensions(required_extensions: &mut xr::ExtensionSet) {
     required_extensions.khr_vulkan_enable = true;
+
+    if cfg!(feature = "editor") {
```
rasmusgo, line 468: I there any reason to use `if cfg!(..` instead of `#[cfg(..`?

hotham-openxr-client/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+pub unsafe extern "system" fn get_instance_proc_addr(
+    _instance: Instance,
+    name: *const c_char,
+    function: *mut Option<pfn::VoidFunction>,
+) -> Result {
+    use std::{ffi::CStr, intrinsics::transmute};
+
+    let name = CStr::from_ptr(name);
+    let name = name.to_bytes();
+    if name == b"xrGetInstanceProcAddr" {
```
rasmusgo, line 26: Could this be a match expression instead?

hotham-editor/src/main.rs
kanerogers: Nope! This can be done on each platform, but each platform has its own unique way of doing it. Focussing on Windows for the first pass.

hotham/src/contexts/xr_context/mod.rs
kanerogers: Can't do attributes on statements, apparently. 🤷

hotham-openxr-client/src/lib.rs
kanerogers: Eh. I've tried a couple different implementations of this - the if statements are messy and yuck, but they work.

.vscode/settings.json
```diff
@@ -6,5 +6,9 @@
     "files.insertFinalNewline": true,
     "files.trimFinalNewlines": true,
     "files.trimTrailingWhitespace": true,
-    "webgl-glsl-editor.format.placeSpaceAfterKeywords": true
+    "webgl-glsl-editor.format.placeSpaceAfterKeywords": true,
+    "rust-analyzer.check.extraArgs": [
+        "--target-dir",
+        "C:/windows/temp/rust-analyzer-check"
+    ],
```
rasmusgo, line 13: This doesn't look portable

hotham-editor/src/main.rs
```diff
@@ -0,0 +1,418 @@
...
+    event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
+    event_loop::ControlFlow,
+    platform::run_return::EventLoopExtRunReturn,
+};
+
+use crate::camera::Camera;
+
+/// Compile your own damn shaders! LazyVulkan is just as lazy as you are!
+static FRAGMENT_SHADER: &'_ [u8] = include_bytes!("shaders/triangle.frag.spv");
+static VERTEX_SHADER: &'_ [u8] = include_bytes!("shaders/triangle.vert.spv");
```
rasmusgo, line 27: These files are not checked in

kanerogers: Ah! Good pickup.

.vscode/settings.json
kanerogers: Yeah, I'm thinking I should just remove this file from the repo

rasmusgo: I'm not convinced the best way forward is to remove the whole file. These settings are good for keeping the files tidy and consistent. Can we at least resolve this in a separate PR?

hotham-editor/src/main.rs
rasmusgo: You may want to add a `.gitignore` with exceptions for these files. Checking in files that are ignored by git can lead to confusing stuff, eg. local changes are not picked up by git even though the file is tracked.

.vscode/settings.json
kanerogers: I guess the problem is that there's always going to be a case where a developer might have overrides that are unique to their specific environment, right? For me the rust analyzer one is super common that I need *per project* to avoid locking issues.

I think anything related to style etc should probably best be left to clippy, otherwise there's no way to, for example. get developers who aren't using VSCode to have their code correctly formatted.

hotham-editor/src/main.rs
kanerogers: Hrm, I've found that it's usually fine after you've explicitly added that file to the index, but I've added an exception to the gitignore just to be on the safe side

hotham-editor/src/shaders/triangle.vert
```diff
@@ -0,0 +1,14 @@
+#version 450
+
+layout(location = 0) in vec4 in_position;
+layout(location = 1) in vec4 in_colour;
```
rasmusgo, line 4: "color"

kanerogers: scowl

kanerogers:
Need to fix the dependencies so we're not pulling in UDS on Linux.

hotham-openxr-client/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+    } else if name == b"xrGetActionStateBoolean" {
+        *function = transmute::<pfn::GetActionStateBoolean, _>(get_action_state_boolean);
+    } else if name == b"xrApplyHapticFeedback" {
+        *function = transmute::<pfn::ApplyHapticFeedback, _>(apply_haptic_feedback);
+    } else if name == b"xrEndSession" {
+        *function = transmute::<pfn::EndSession, _>(end_session);
+    } else {
+        let _name = String::from_utf8_unchecked(name.to_vec());
+        unsafe extern "system" fn bang() -> Result {
+            panic!("UNIMPLEMENTED FUNCTION!");
```
louisgjohnson, line 140: Could potentially add the name to the panic to make it easier to track down.

kanerogers: This is a fun one! So I've tried to pass this in but for whatever reason - potentially FFI weirdness, I am clearly not clever enough to understand those machinations - it isn't possible to return a function that will panic with the correct function name.

PR #422, opened 2023-02-10
Rasmusgo/hot reload custom shaders

rasmusgo:
This improves the system for hot reloading by exposing a list of recently updated files and allowing custom shaders to be stored in the engine and updated by the hot reload system. The engine doesn't know how to use the shaders so the app must handle that itself. This is done by the custom rendering example.

### DIFF
--- a/hotham/src/contexts/render_context.rs
+++ b/hotham/src/contexts/render_context.rs
@@ -44,6 +44,9 @@ use glam::{Affine3A, Mat4, Vec3, Vec4};
 use openxr as xr;
 use vk_shader_macros::include_glsl;
 
+static VERT_NAME: &str = "pbr.vert.spv";
+static FRAG_NAME: &str = "pbr.frag.spv";
+static COMPUTE_NAME: &str = "culling.comp";
 static VERT: &[u32] = include_glsl!("src/shaders/pbr.vert", target: vulkan1_1);
 static FRAG: &[u32] = include_glsl!("src/shaders/pbr.frag", target: vulkan1_1);
 static COMPUTE: &[u32] = include_glsl!("src/shaders/culling.comp", target: vulkan1_1);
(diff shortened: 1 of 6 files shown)


hotham/src/contexts/render_context.rs
```diff
@@ -134,7 +109,11 @@ impl RenderContext {
         let pipeline_layout =
             create_pipeline_layout(vulkan_context, slice_from_ref(&descriptors.graphics_layout))?;
 
-        let shaders = Default::default();
+        let shaders = Shaders::from([
+            (VERT_NAME.into(), VERT.into()),
+            (FRAG_NAME.into(), FRAG.into()),
+            (COMPUTE_NAME.into(), COMPUTE.into()),
```
rasmusgo, line 115: The compute shader is created with `COMPUTE` directly in `create_compute_pipeline`. What do you think about inlining the shaders here to avoid them being used elsewhere by mistake?

hotham/src/contexts/render_context.rs
```diff
@@ -44,6 +44,9 @@ use glam::{Affine3A, Mat4, Vec3, Vec4};
 use openxr as xr;
 use vk_shader_macros::include_glsl;
 
+static VERT_NAME: &str = "pbr.vert.spv";
+static FRAG_NAME: &str = "pbr.frag.spv";
+static COMPUTE_NAME: &str = "culling.comp";
```
kanerogers, line 49: Need a ".spv" here for the compute shader

hotham/src/contexts/render_context.rs
```diff
@@ -71,35 +74,7 @@ pub struct RenderContext {
...
-        compute_shader: Vec<u32>,
-    ) -> Self {
-        Self {
-            vertex_shader,
-            fragment_shader,
-            compute_shader,
-        }
-    }
-}
+pub type Shaders = HashMap<String, Vec<u32>>;
```
kanerogers, line 77: I think we lose something by getting rid of the struct here! `RenderContext` has *only* three shaders that all do very specific things - having an arbitrary number of them, identified by a string doesn't seem to meet that goal.

hotham/src/contexts/render_context.rs
```diff
@@ -71,35 +74,7 @@ pub struct RenderContext {
     pub primitive_map: HashMap<u32, InstancedPrimitive>,
 }
 
-pub struct Shaders {
-    pub vertex_shader: Vec<u32>,
-    pub fragment_shader: Vec<u32>,
-    pub compute_shader: Vec<u32>,
-}
-
-impl Default for Shaders {
```
kanerogers, line 80: This default was here for a reason. 😛 `RenderContext` is not designed to be a generic renderer - it has very specific knowledge about the shaders and pipeline it's using to render them, so it makes sense to have strong defaults.

hotham/src/contexts/render_context.rs
kanerogers: I'm not sure what you mean by this - can you elaborate?

rasmusgo: I was thinking something like

```rs
            (VERT_NAME.into(), include_glsl!("src/shaders/pbr.vert", target: vulkan1_1)),
            (FRAG_NAME.into(), include_glsl!("src/shaders/pbr.frag", target: vulkan1_1)),
            (COMPUTE_NAME.into(), include_glsl!("src/shaders/culling.comp", target: vulkan1_1)),
```

But it got messy with casting now that I tried it.

rasmusgo:
Closed in favor of #424.

PR #424, opened 2023-02-10
Rasmusgo/hot reload custom shaders 2

rasmusgo:
This PR replaces #422.

All assets which are received by the hot reloading are exposed by `engine.get_updated_assets()`. This enables the custom rendering example to look for its shaders and update them. It could also be used for other assets in the future.


kanerogers (approved):
Sorry for taking a while to come back to this!

Looking good - thanks for the improvements - this is definitely a step in the right direction!

PR #425, opened 2023-02-18
Fix dummy keystore

rasmusgo:
Olly reported a problem with the dummy keystore:
> Hi!  Was able to get hotham working on my setup, but had the following minor issue, that others might encounter.  I was building from an OSX environment and the dummy keystore was hilighted as invalid when I tried to run the build on my device.  I think this is because the ./examples/hotham_examples.keystore file was checked in as text instead of binary.  I regenerated a keystore locally and was able to upload the apk, but you might want to mark the file as binary, so that others won't hit this problem from an "out of the box" experience.

> I think adding the following to .gitattributes should do the trick:
`*.keystore binary`

This is an attempt at the suggested solution.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,3 +9,4 @@ test_assets/culling_stress_test.glb filter=lfs diff=lfs merge=lfs -text
 hotham/data/brdf_lut.ktx2 filter=lfs diff=lfs merge=lfs -text
 hotham/data/environment_map_diffuse.ktx2 filter=lfs diff=lfs merge=lfs -text
 hotham/data/environment_map_specular.ktx2 filter=lfs diff=lfs merge=lfs -text
+*.keystore binary


oliy:
This looks good.  Incidentally, my actual problem turned out to be that OSX installs/uses Java 8 by default.  The actual fix for my problem was to install a modern Java runtime (i.e. using `brew install java`), then pointing `JAVA_HOME` to the correct runtime before running any of the examples.

rasmusgo:
Git already considers this file binary:
```sh
$ git diff --numstat 4b825dc642cb6eb9a060e54bf8d69288fbee4904 HEAD -- | grep "^-" | cut -f 3
docs/hotham_sync_loop.png
examples/complex-scene/openxr_loader.dll
examples/crab-saber/openxr_loader.dll
examples/custom-rendering/openxr_loader.dll
examples/hotham_examples.keystore
examples/simple-scene/openxr_loader.dll
hotham-simulator/build_input/viewdisplay.frag.spv
hotham-simulator/build_input/viewdisplay.vert.spv
hotham-simulator/openxr_simulator.reg
hotham-simulator/src/shaders/viewdisplay.frag.spv
hotham-simulator/src/shaders/viewdisplay.vert.spv
logo.jpg
openxr_loader.dll
test_assets/ferris-the-crab/textures/ferris_eyes_0.png
```
(command taken from [here](https://stackoverflow.com/a/66289356/302951))

There is no need to explicitly mark it as binary.

PR #436, opened 2023-03-31
Rasmusgo/upgrade deps

rasmusgo:
Upgrade some dependencies. I reverted the quinn upgrade because it needs a bit more work and that can go in another PR.


kanerogers:
LGTM!

PR #437, opened 2023-04-07
Rasmusgo/xr time now

rasmusgo:
This PR adds a method that retrieves the current time as a `openxr::Time` that can be passed to openxr instead of predicted display time to get more accurate poses with the trade-off that they will be old when the frame is done rendering and displayed to the user.

This function relies on the appropriate time conversion extension being enabled. It can be enabled when creating the engine like this:

```rs
    let mut extensions = xr::ExtensionSet::default();
    #[cfg(windows)]
    {
        extensions.khr_win32_convert_performance_counter_time = true;
    }
    #[cfg(not(windows))]
    {
        extensions.khr_convert_timespec_time = true;
    }
    let extensions = Some(extensions);
    let mut engine_builder = EngineBuilder::new();
    engine_builder.openxr_extensions(extensions);
    let mut engine = engine_builder.build();
```
_(I couldn't get it prettier because of moving mut and temporary variables being dropped. The builder pattern could probably be made to work better.)_

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,7 @@
     "files.insertFinalNewline": true,
     "files.trimFinalNewlines": true,
     "files.trimTrailingWhitespace": true,
+    // Uncomment this to tell rust-analyzer to treat the code as being compiled for android.
+    // "rust-analyzer.cargo.target": "aarch64-linux-android",
     "webgl-glsl-editor.format.placeSpaceAfterKeywords": true
 }
--- a/hotham/Cargo.toml
+++ b/hotham/Cargo.toml
@@ -51,3 +51,6 @@ serde_json = "1.0"
 jni = "0.19.0"
 ndk = "0.6"
 ndk-glue = "0.6"
+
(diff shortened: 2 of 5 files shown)


kanerogers:
How do you feel about just enabling the extension regardless? Doesn't seem like there'd be much of an issue doing that?

rasmusgo:
> How do you feel about just enabling the extension regardless? Doesn't seem like there'd be much of an issue doing that?

Sounds good to me! I'll try to find a good place to put it.

rasmusgo:
Enabling it automatically breaks the compatibility with the hotham simulator:
```
Error [GENERAL | xrCreateInstance | OpenXR-Loader] : LoaderInstance::CreateInstance, no support found for requested extension: XR_KHR_win32_convert_performance_counter_time
Error [GENERAL | xrCreateInstance | OpenXR-Loader] : xrCreateInstance failed
thread 'main' panicked at '!!FATAL ERROR - Unable to initialize OpenXR!!: a requested extension is not supported', hotham\src\engine.rs:106:14
```

Issue #434, opened 2023-03-10
[Bug]: Textures mixed up between different materials


rasmusgo:
### What happened?

There is something wrong with how textures are indexed when some materials don't have textures. The material of the helmet and the cube has gotten mixed up somehow. The helmet should have textures and the cube should not. The screenshot is with `[HOTHAM_DEBUG] params.z is now 3`
![OculusScreenshot1678396415](https://user-images.githubusercontent.com/1162652/224378957-a4e4a61c-42e9-43bc-8f6f-c87a9d18dd92.jpeg)

### Version

latest

### In which Hotham component are you seeing the problem on?

Core

### What VR System are you seeing the problem on?

None

### What OS are you seeing the problem on?

Windows

PR #438, opened 2023-04-09
Rasmusgo/floor plane

rasmusgo:
This adds a floor to the simple scene example and fixes the transform of the helmet.

Incidentally, this also serves as a reproduction of https://github.com/leetvr/hotham/issues/434.

image

### DIFF
--- a/hotham/src/lib.rs
+++ b/hotham/src/lib.rs
@@ -23,6 +23,7 @@ pub use glam;
 pub use hecs;
 pub use hotham_error::HothamError;
 pub use id_arena;
+pub use rapier3d::na;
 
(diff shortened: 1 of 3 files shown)


hotham/src/lib.rs
```diff
@@ -23,6 +23,7 @@ pub use glam;
 pub use hecs;
 pub use hotham_error::HothamError;
 pub use id_arena;
+pub use rapier3d::na;
```
rasmusgo, line 26: This is needed for
```rs
let collider = Collider::new(SharedShape::halfspace(na::Vector3::y_axis()));
```

kanerogers (approved):
Nice one! Thanks!

PR #441, opened 2023-08-22
Rasmusgo/parent hierarchy

### LINKED ISSUE #435: Do parent hierarchy less dumb
It turns out that there's no reason to recreate the graph when we already have one; traverse bottom up instead of top down.

rasmusgo:
* Put transformation from child to parent inside `Parent` component and remove `LocalTransform` component.
* Simplify computation of `GlobalTransform` based on `hecs` example code.

This resolves #435.


examples/complex-scene/src/lib.rs
```diff
@@ -108,9 +107,13 @@ fn add_helmet(
...
-        let mut local_transform = world.get::<&mut LocalTransform>(helmet).unwrap();
-        local_transform.translation = translation;
-        local_transform.scale = [0.5, 0.5, 0.5].into();
+        let mut global_transform = world.get::<&mut GlobalTransform>(helmet).unwrap();
+        let (_, rotation, _) = global_transform.0.to_scale_rotation_translation();
+        global_transform.0 = glam::Affine3A::from_scale_rotation_translation(
+            [0.5, 0.5, 0.5].into(),
+            rotation,
+            translation,
+        );
```
rasmusgo, line 116: This is an example that could be simplified by enforcing uniform scaling. However, `glam` does not have primitives for enforcing uniform scaling ("similarity transforms" aka "conformal") which makes it less worth pursuing. There is a long discussion thread in the `glam` repo which [ends in sticking with only the affine transformation type](https://github.com/bitshifter/glam-rs/issues/25#issuecomment-1170979733).

PR #442, opened 2023-08-23
rasmusgo/parent hierarchy 2

### LINKED ISSUE #435: Do parent hierarchy less dumb

rasmusgo:
Resolves #435 without removing the `LocalTransform` component. Supersedes #441.


kanerogers (approved):
This looks a lot better to me. Yeah, there's still a bit of annoying redundancy with the Local/Global transform split but I think we can live with that for now.

PR #443, opened 2023-08-25
Rasmusgo/quadrics gallery

rasmusgo:
OculusScreenshot1692956023

* Improve grabbing system by preserving relative pose to grabbed object.
* Add quadrics that can be manipulated by grabbing and moving control points.
* Render hands with lighting. (instead of unlit).
* Improve quadrics shader to reduce artifacts.
* Run clippy on tests in CI.


kanerogers (approved):
Looking fantastic!! Nice work.

